PR #66, opened 2021-03-24
Hotfix - SaveLoad Configs

rvalim:
### What was happening
* Load view isn't working on production

### What was done
* The column.binding is back to the columns Map
* Created new methods _getColumns(): []_, _hasColumn(key: string)_
* To avoid the usage of grid.columns as an array, the getter was removed, and the methods _hasColumn_, _getColumn_ and _getColumns_ should be used on its place

### Test Steps
1. Test the Validations
2. Test the Save/Load View

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/Features/ValidationMark.ts
+++ b/code/src/Features/ValidationMark.ts
@@ -410,7 +410,7 @@ namespace Features {
          */
         public validateRow(rowNumber: number): void {
             // Triggers the validation method per column
-            this._grid.columns.forEach((column: Column.IColumn) => {
+            this._grid.getColumns().forEach((column: Column.IColumn) => {
                 // This method gets executed by an API. No values change in columns, so the current value and the original one (old value) are the same.
                 const currValue = this._grid.provider.getCellData(
                     rowNumber,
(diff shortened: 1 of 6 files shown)


glundgren93 (approved):

code/src/Features/ValidationMark.ts
```diff
@@ -410,7 +410,7 @@ namespace Features {
          */
         public validateRow(rowNumber: number): void {
             // Triggers the validation method per column
-            this._grid.columns.forEach((column: Column.IColumn) => {
+            this._grid.getColumns().forEach((column: Column.IColumn) => {
```
rugoncalves, line 413: Wouldn't it be better to be a getter? (and keep the same name as before?)

code/src/Grid/AbstractGrid.ts
```diff
@@ -158,21 +170,22 @@ namespace Grid {
...
+                return this.getColumns().find((p) => p && p.equalsToID(key));
             }
         }
 
+        public getColumns(): Column.IColumn[] {
+            return Array.from(this._columnsSet);
+        }
+
+        public hasColumn(key: string): boolean {
+            return this._columns.has(key);
```
rugoncalves, line 186: Shouldn't this method, look for the other ID's as well? or should we keep this case for a future improvement?

rugoncalves (approved):

PR #69, opened 2021-03-25
Hotfix for width and visibility during Save/Load View

rvalim:
### What was happening
* Width wasn't being reseted when width = undefined
* Visible wasn't being reseted when visible = undefined

### What was done
* Placing defaults for width and visible, null and true respectively

### Checklist
* [x] tested locally
* [x] documented the code


rugoncalves (approved):
Looks good

tmbp95 (approved):

PR #70, opened 2021-03-25
RGRIDT-731: Fixing Regex

glundgren93:
This PR is for a bug that happened in Safari. 

### What was happening
* We were using look behind in our regex. This is not supported in safari, so it was throwing an error whenever we set a new language to grid.

### What was done
*  Changed regex.

### Screenshots
safari

### Checklist
* [x] tested locally
* [x] documented the code


rugoncalves (approved):
Well caught!

rvalim (dismissed):

tmbp95 (approved):

tmbp95 (approved):

rvalim (approved):

PR #74, opened 2021-04-15
RGRIDT-748: DataSource review

rvalim:
Datasource encapsulation, more details on Jira.

### What was happening / What was done
* Treatment for ToOSFormat or ToJSONFormat spread by the code
* FlexGrid class, responsable for rendering the FlexGrid were handling data
* GridAPI.Filter had wijmo code, that was sent to the new DataSource, and maybe will be removed on the future, in case of using wijmo search engine
* Adding and Deleting rows were manipulating the dataset, makes sense this logic be handled by the DataSource also, so code was moved.
* If we create a TransposedGrid for example, the logic for autoGenerateColumns needs to be replicated. For example, 
  * Does the user uses the ArrangeData? If so, trigger the ColumnGenerator
  * If not, flatten the data and them call the provider autoGenerate
  * This kind of treatment is now handled by the AbstractGrid, and setting data will be done by DataSource

### Test Steps
1. Navigate throw the pages, verify if everything is working fine

### Checklist
* [x] tested locally
* [ ] documented the code


code/src/OSFramework/Helper/JSONParser.ts
```diff
@@ -63,56 +21,4 @@ namespace OSFramework.Helper {
             .toISOString()
             .substr(0, 10);
     }
```
tmbp95, line 23: These 2 methods will stay here?

code/src/OSFramework/Interface/IDataSource.ts
```diff
@@ -0,0 +1,73 @@
...
+         * Return the provider's dataSource
+         */
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        getProviderDataSource(): any;
+        /**
+         * Indicate whether the filtered dataSource has Results to show
+         */
+        hasResults(): boolean;
+        /**
+         * Responsabem to items from the datasource
```
tmbp95, line 58: can you fix the comment pls?

code/src/OSFramework/Interface/IDataSource.ts
```diff
@@ -0,0 +1,73 @@
+namespace OSFramework.Grid {
+    /**
+     * Defines the basic interface for a DataSource
+     */
+    export interface IDataSource extends OSFramework.Interface.IBuilder {
+        /**
+         * Idenfity if metadata was given
```
tmbp95, line 7: can you fix the comment pls?

code/src/OSFramework/OSStructure/GridChanges.ts
```diff
@@ -0,0 +1,8 @@
+/// <reference path="../../OSFramework/OSStructure/changesDone.ts" />
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+namespace OSFramework.OSStructure {
+    export class GridChanges extends OSFramework.OSStructure.ChangesDone {
+        hasInvalidLines: boolean;
+        invalidLinesJSON: string;
```
tmbp95, line 6: TOP!

code/src/OSFramework/Grid/AbstractDataSource.ts
```diff
@@ -0,0 +1,222 @@
...
+     * @param data Data to format
+     */
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    function ToOSFormat(
+        convertions: Map<string, Set<string>>,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        data: Array<any>
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    ): void {
+        //TODO: [RGRIDT-638] Regression 2021-02-12: Is this method the best solution
```
rugoncalves, line 82: Maybe... just maybe... let's no leave this comment here!?! :D

code/src/OSFramework/Grid/AbstractDataSource.ts
```diff
@@ -0,0 +1,222 @@
...
+                });
+            };
+
+            converter(parsedNewItem);
+
+            return parsedNewItem;
+        }
+
+        public get hasMetadata(): boolean {
+            return !!this._metadata;
```
rugoncalves, line 153: Let's use a more explicit tests, please.

code/src/OSFramework/Grid/AbstractDataSource.ts
```diff
@@ -0,0 +1,222 @@
...
+namespace OSFramework.Grid {
+    /**
+     * Used to flatten the data array
+     * @param dataArray
+     * @returns
+     */
+    function FlattenArray(dataArray: JSON[]): JSON[] {
+        const returnDataArray = [];
+        dataArray.forEach((item) => {
+            returnDataArray.push(OSFramework.Helper.Flatten(item));
```
rugoncalves, line 11: Why wasn't the Flatten function moved here as well? I don't think it is used elsewhere...

code/src/OSFramework/Grid/AbstractDataSource.ts
```diff
@@ -0,0 +1,222 @@
...
+            return this._metadata;
+        }
+
+        public removeRow(item: number | JSON): boolean {
+            const index =
+                typeof item === 'number' ? item : this._ds.indexOf(item);
+
+            if (index === -1) return false;
+
+            this._ds.splice(index, 1);
```
rugoncalves, line 184: What happens when the index is out of boundaries?
Maybe we should protect this code a bit more.

code/src/OSFramework/Grid/AbstractGrid.ts
```diff
@@ -179,25 +200,88 @@ namespace OSFramework.Grid {
...
+                            throw `The binding "${
+                                column.config.binding
+                            }" doesn't match any valid column from the data you specified. ${'\n'} Expected format: "EntityName.FieldName". ${'\n'} For example: "Product_Sample.Name"`;
+                        }
+                        // If keyword is a property from metadata then use metadata[keyword] as the new metadata and iterate to the next keyword.
+                        metadata = metadata[keyword];
+                    });
+                });
+            }
+        }
```
rugoncalves, line 285: Private actions should be before public ones.

code/src/OSFramework/OSStructure/GridChanges.ts
rugoncalves: Why create a second class? and not add these fields to the class ChangesDone?

rugoncalves (commented):
Good job!

code/src/OSFramework/OSStructure/GridChanges.ts
rvalim: this is a feature data... invalidLines comes from ValidationFeature...

DataSource treats only changes, deletion and insertion....

This second class can be incremented without affecting DataSource

code/src/OSFramework/Helper/JSONParser.ts
rvalim: They are being used by the ValidationMark. Does it make sense?

code/src/OSFramework/Grid/AbstractDataSource.ts
rvalim: This //TODO was in the code already... could you fix it for me? =D

code/src/OSFramework/Grid/AbstractDataSource.ts
rvalim: I don't move it because the ActionColumn is using it... can you tell me if it is really needed?

rugoncalves: At the moment yes. The Action column is using it to flatten the line data to trigger the event. Do you think that this should be done in a different way?

code/src/OSFramework/Grid/AbstractDataSource.ts
tmbp95: const months1 = ['Jan', 'March', 'April', 'June'];
months1.splice(1, 1, 'May');
console.log(months1);
// Array ["Jan", "Maasdfasdy", "April", "June"]

const months2 = ['Jan', 'March', 'April', 'June'];
months2.splice(123523512, 1, 'May');
console.log(months2);
// Array ["Jan", "March", "April", "June", "May"]

tmbp95: It seems ok to me. @rugoncalves

code/src/OSFramework/Grid/AbstractDataSource.ts
rugoncalves: FlattenArray is only used, by the DataSource, hence the reason to move it here, and make it private.
Regarding the Flatten function itself, it is used by the Action Column, so it should be in common place.

code/src/OSFramework/Grid/AbstractDataSource.ts
rugoncalves: perfect

rugoncalves (dismissed):

tmbp95 (dismissed):

tmbp95 (approved):

rugoncalves (approved):
Thank you @rvalim and @tmbp95! Good job!

PR #80, opened 2021-04-30
RGRIDT-770: Check if column exists before adding them again

glundgren93:
### What was happening
* Whenever we used server-side search, the columns were being duplicated

### What was done
* Added a condition before the step of adding columns

### Test Steps
1. Open sample
2. On server-side search, search for pink
3. Check if new columns were not added. 

### Screenshots
(prefer animated gif)

### Checklist
* [ ] tested locally
* [ ] documented the code

### DIFF
--- a/code/src/OSFramework/Grid/AbstractGrid.ts
+++ b/code/src/OSFramework/Grid/AbstractGrid.ts
@@ -249,7 +249,9 @@ namespace OSFramework.Grid {
                     this.dataSource.getMetadata(),
                     this.config.allowEdit
                 );
-                generated.forEach((p) => this.addColumn(p));
+                if (this._columns.size === 0) {
+                    generated.forEach((p) => this.addColumn(p));
+                }
             } else {
                 //if the grid is read-only, then we'll flatten the array and use wijmo generator
                 if (!this.config.allowEdit) {


code/src/OSFramework/Grid/AbstractGrid.ts
```diff
@@ -249,7 +249,9 @@ namespace OSFramework.Grid {
                     this.dataSource.getMetadata(),
                     this.config.allowEdit
                 );
-                generated.forEach((p) => this.addColumn(p));
+                if (this._columns.size === 0) {
+                    generated.forEach((p) => this.addColumn(p));
+                }
```
rugoncalves, line 254: Does this cover the scenario, when a new column is added as well?

bmartinho: I was checking the same and this code only runs in auto generated columns that have metadata (where created with ArrangeData), so no problem with add/remove columns

rugoncalves: As discussed, we will leave for now as it is. @glundgren93 if you can add just an improvement to backlog, to (maybe) support swapping data sources in the future.

rugoncalves (approved):

code/src/OSFramework/Grid/AbstractGrid.ts
bmartinho: yes!

bmartinho (approved):
Let's go!

tmbp95 (approved):

PR #89, opened 2021-05-31
RGRIDT-832: Fixing export excel with styles

glundgren93:
This PR fixes conditional format not being applied on excel export.

### What was happening
* Conditional format was not being applied on excel export.

### What was done
* Adjusted exportToExcel method in order to apply conditional format on all cells

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/Export.ts
+++ b/code/src/WijmoProvider/Features/Export.ts
@@ -82,24 +82,22 @@ namespace WijmoProvider.Feature {
 
         public exportToExcel(withStyles: boolean): void {
             this._resetPagination();
-
-            const params = {
-                includeColumnHeaders: true,
-                includeRowHeaders: true,
-                includeCellStyles: withStyles,
-                formatItem: this.exportFormatItem
-            };
-
-            const book = wijmo.grid.xlsx.FlexGridXlsxConverter.save(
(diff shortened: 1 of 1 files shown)


code/src/WijmoProvider/Features/Export.ts
```diff
@@ -82,24 +82,22 @@ namespace WijmoProvider.Feature {
...
-                params
-            );
-            //name the sheet
-            book.sheets[0].name = 'DataGrid Data';
-            // save the book
-            book.save('DataGridReactive.xlsx');
-
-            this._reApplyPagination();
+            // include timeout in order to apply conditional format
+            setTimeout(() => {
```
tmbp95, line 86: We were doing this exact same setTimeout for the trad web, right?

glundgren93: yea

tmbp95 (approved):
I wish we were able to avoid the setTimout :(

code/src/WijmoProvider/Features/Export.ts
bmartinho: Let's keep this also in Reactive

bmartinho (approved):
Ready to go

PR #92, opened 2021-06-02
Stories/rgridt 687

JoaoFerreira-FrontEnd:
This PR is for...

[https://pp-datagridteam-dev.outsystemsenterprise.com/DataGridSampleReactive_RGRIDT687/SaveChanges](url)

### What was happening

### What was done

### Test Steps
2. Change the validation status, _IsValid_ value to False;
3. On another button use a new action that uses **MarkChangesAsSaved** client action;
4. Config the value of _ClearInvalids_ to **_False_**
5. Call this action after changing the value on the previous column
7. Config the value of _ClearInvalids_ to **_True_**
8. Repeat the test

### Screenshots
(prefer animated gif)

### Checklist
* [x] tested locally
* [x] documented the code


glundgren93:
Nice work! For the next pull requests, please create your branch following our convention: RGRIDT-XXX. This will guarantee that our pipeline triggers our automated tests and deploy on your OS branch.

PR #93, opened 2021-06-02
RGRIDT-687: Implement MarkChangesAsSaved behavior with invalid cells

JoaoFerreira-FrontEnd:
### What was happening

### What was done

### Test Steps
2. Change the validation status, IsValid value to False;
3. On another button use a new action that uses MarkChangesAsSaved client action;
4. Config the value of ClearInvalids to False;
5. Call this action after changing the value on the previous column;
7. Config the value of ClearInvalids to True;
8. Repeat the test;

### Checklist
* [x] tested locally
* [x] documented the code


tmbp95 (approved):
Nice one!

glundgren93 (dismissed):

code/src/GridAPI/GridManager.ts
```diff
@@ -145,13 +145,17 @@ namespace GridAPI.GridManager {
      *
      * @export
      * @param {string} gridID ID of the Grid where the change will occur.
+     * @param {boolean} clearInvalids determines whether or not we should clean the validation marks.
      */
-    export function MarkChangesAsSaved(gridID: string): void {
+    export function MarkChangesAsSaved(
+        gridID: string,
+        clearInvalids: boolean
```
rugoncalves, line 152: This should have a default value...

rugoncalves (changes requested):

rugoncalves (approved):
Looking good!

tmbp95 (approved):

code/src/OSFramework/Feature/IValidationMark.ts
```diff
@@ -5,7 +5,8 @@ namespace OSFramework.Feature {
         invalidRows: Array<any>;
         clear(): void;
         errorMessage(rowNumber: number, binding: string): string;
-        isInvalid(rowNumber: number, binding: string): boolean;
+        isInvalidByRow(row: any): boolean;
+        isInvalidByRowNumber(rowNumber: number, binding: string): boolean;
```
rugoncalves, line 9: I would suggest to keep it as it was. Add only the **isInvalidRow** method.

code/src/OSFramework/Grid/IGrid.ts
```diff
@@ -29,7 +29,7 @@ namespace OSFramework.Grid {
         ): void;
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         changeProperty(propertyName: string, propertyValue: any): void;
-        clearAllChanges(): void;
+        clearAllChanges(clearInvalids: boolean): void;
```
rugoncalves, line 32: I think a better name could be, _forceClearDirtyMarks_ what do you think?

code/src/WijmoProvider/Features/DirtyMark.ts
```diff
@@ -164,22 +165,22 @@ namespace WijmoProvider.Feature {
             this._grid.provider.invalidate(); //Mark to be refreshed
         }
 
-        // public clearByRow(row: number): void {
-        //     this._metadata.clearPropertyByRow(row, this._internalLabel);
-        //     this._grid.grid.invalidate(); //Mark to be refreshed
-        // }
+        public clearByRow(row: any): void {
+            this._metadata.clearPropertyByRow(row, this._internalLabel);
```
rugoncalves, line 169: clearPropertyInRow would be more accurate...

rugoncalves (commented):
Other than this good!
I'm thinking since this branch has soo many commits if wouldn't be better to cherry pick the commits or to create a new branch and implement simply the differences.

code/src/WijmoProvider/Grid/RowMetadata.ts
```diff
@@ -56,9 +56,19 @@ namespace WijmoProvider.Grid {
             });
         }
 
-        public clearPropertyByRow(row: number, propertyName: string): void {
-            this.hasOwnProperty(row, propertyName) &&
-                this._getRowMetadata(row).delete(propertyName);
+        public clearPropertyByRow(dataItem: any, propertyName: string): void {
+            // eslint-disable-next-line prettier/prettier
+            if(dataItem.__osRowMetada && dataItem.__osRowMetada.has(propertyName)){
+                dataItem.__osRowMetada.delete(propertyName);
```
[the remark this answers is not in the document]
JoaoFerreira-FrontEnd: Got it!

code/src/OSFramework/Feature/IValidationMark.ts
JoaoFerreira-FrontEnd: Here, should be like this **isInvalid(rowNumber: number, binding: string): boolean** (original method) and add **isInvalidRow(row: any): boolean** (new need)?

code/src/OSFramework/Grid/IGrid.ts
JoaoFerreira-FrontEnd: We refer these as ValidationMarks but they are at the same time DirtyMarks.
I think changing the name to  forceClearValidationMarks, probably will be better.

PR #97, opened 2021-06-18
RGRIDT-687: MarkChangesAsSaved behavior with invalid cells;

JoaoFerreira-FrontEnd:
### What was happening

### What was done

### Test Steps
2. Change the validation status, IsValid value to False;
3. On another button use a new action that uses MarkChangesAsSaved client action;
4. Config the value of ClearInvalids to False;
5. Call this action after changing the value on the previous column;
7. Config the value of ClearInvalids to True;
8. Repeat the test;

### Checklist
* [x] tested locally
* [x] documented the code


code/src/OSFramework/Interface/IRowMetadata.ts
```diff
@@ -22,31 +22,64 @@ namespace OSFramework.Interface {
...
-        clearPropertyByRow(row: number, propertyName: string): void;
+        clearPropertyByRowNumber(row: number, propertyName: string): void;
         /**
          * Responsible for retrieving metadata information for a given row and property
          * @param row Index row reference, works only in the current page
          * @param propertyName Metadata property
          */
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        getMetadata(row: number, propertyName: string): any;
+        getMetadataByRow(row: any, propertyName: string): any;
```
rugoncalves, line 42: I would change the name to _getMetadataInRow_. This is a more direct name :)

rugoncalves (approved):
Minor name reviews if you think will look better. :) 
Congratulations for all this work 🥇

rugoncalves (approved):

code/src/OSFramework/Feature/IDirtyMark.ts
```diff
@@ -3,7 +3,7 @@ namespace OSFramework.Feature {
     export interface IDirtyMark {
         clear(): void;
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        clearPropertyInRow(row: any): void;
```
tmbp95, line 6: Can you please add  a comment here? We should have comments all over the interfaces to make it easier for the other developers.

tmbp95 (approved):

bmartinho (approved):
LGTM

PR #98, opened 2021-06-22
RGRIDT-856: Added logic to enable set a column as not hidable

rugoncalves:
This PR is for adding a configuration to enable the developer to set a column as not hidable in run time by the end user.
Based on the [PR#90](https://github.com/OutSystems/outsystems-datagrid-reactive/pull/90) made by [@rmadeiraneto](https://github.com/rmadeiraneto).

### What was happening
* There was no lowcode way for the developer to make a column not hidable in runtime by the end-user.

### What was done
* Added a new boolean property to the ColumnConfig class
* Removed property not in use in ColumnConfig class
* Added formatItem callback to the ColumnPicker class where the new property will be checked

### Test Steps
1. Set the OptionalConfig CanBeHidden to false in a column
2. In runtime, open the Column Picker
2. Notice that the column will not be hidable

### Screenshots
image

### Checklist
* [x] tested locally
* [x] documented the code
* [x] UML updated


code/src/OSFramework/Configuration/Column/ColumnConfig.ts
```diff
@@ -25,7 +27,6 @@ namespace OSFramework.Configuration.Column {
         public header: string;
         public isMandatory: boolean;
         public multiLine: boolean;
-        public required: boolean;
```
tmbp95, line 28: Be careful when you remove this property, if I remember correctly, it has a direct impact on the deletion of the cell's content. So if you have the required = true, you won't be able to delete the content of any cell from that column. If you set it to false, you will delete the content of the cells from that column just by pressing DEL key.

glundgren93 (dismissed):

bmartinho (dismissed):
LGTM

bmartinho (approved):
Ready for merge

PR #99, opened 2021-06-23
RGRIDT-861: Adding the ability to export with a specific file name.

rugoncalves:
This PR is for adding the ability for developers to provide the filename with which the excel/csv file will be exported with.

### What was happening
* The name of the excel/csv was hardcoded in the code

### What was done
* A new input parameter was added to the feature export, name "Filename", with default value "DataGridReactive".
* A new function was created to handle the name properly

### Test Steps
1. Add a filename in the ContextMenu option "Export to Excel"
2. Validate that the file exported has the filename + extension passed in OutSystems
3. Leave the filename field empty
4. Validate that the output file has the filename 'DataGridReactive' + extension.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/Export.ts
+++ b/code/src/WijmoProvider/Features/Export.ts
@@ -1,7 +1,8 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 namespace WijmoProvider.Feature {
     export class Export
-        implements OSFramework.Feature.IExport, OSFramework.Interface.IBuilder {
+        implements OSFramework.Feature.IExport, OSFramework.Interface.IBuilder
+    {
         private _curPage: number;
         private _grid: WijmoProvider.Grid.IGridWijmo;
         private _pageSize: number;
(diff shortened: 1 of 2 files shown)


code/src/WijmoProvider/Features/Export.ts
```diff
@@ -22,6 +23,14 @@ namespace WijmoProvider.Feature {
             );
         }
 
+        private _handleFilename(fileName: string, isCSV = false): string {
+            if (fileName === undefined || fileName === '') {
```
glundgren93, line 27: Wouldn't be better to set the value for fileName as a default value on the parameter itself?

fileName: string = 'DataGridReactive'

rugoncalves: That only works for when the filename is _undefined_. If the name is '' (empty), then the default value would not be used...

tmbp95 (approved):

glundgren93 (approved):

zipzapduo:
Is it ready for download?

glundgren93:
> Is it ready for download? I mean is it already in Outsystems forge?

Not yet. We plan on making a release next week.

PR #100, opened 2021-06-24
RGRIDT-809: Adding calculated columns

glundgren93:
This PR is for adding calculated columns.

[Sample page](url)

### What was done
* Created Calculated column
* Created Calculated field features that handle formulas

### Checklist
* [x] tested locally
* [ ] documented the code


code/src/WijmoProvider/Features/CalculatedField.ts
```diff
@@ -0,0 +1,95 @@
...
+                                    col.columnType ===
+                                        OSFramework.Enum.ColumnType.Number
+                            )
+                            // get binding
+                            .map((col) => col.config.binding)
+                            .indexOf(value) !== -1
+                );
+
+            if (!isValid) {
+                throw `The content of ${header} column is not being displayed because the given formula is not valid. Formula values must be bindings to Currency or Number columns or numeric values written as text.
```
tmbp95, line 67: A good one to be converted to an error code in the future! (near future)

code/src/WijmoProvider/Columns/CalculatedColumn.ts
```diff
@@ -0,0 +1,42 @@
...
+                columnID,
+                new OSFramework.Configuration.Column.ColumnConfigAdditional(
+                    configs,
+                    extraConfig
+                )
+            );
+            this._columnEvents =
+                new OSFramework.Event.Column.ColumnEventsManager(this);
+
+            // set custom binding
```
bmartinho, line 21: Could you please just comment here what is the logic to create the binding name?

bmartinho (commented):
Just add the comment and it is a LGTM!

code/src/WijmoProvider/Features/CalculatedField.ts
bmartinho: Correct, will create a task in JIRA to implement error code logic in Data Grid Component

code/src/WijmoProvider/Columns/CalculatedColumn.ts
glundgren93: done

bmartinho (approved):
LGTM

tmbp95 (approved):

PR #101, opened 2021-06-24
RGRIDT 849 - Defining column dependency - Other column types

JoaoFerreira-FrontEnd:
### What was happening
*  Wasn't possible to bind an interaction of a cell with the content of another one in the same row.

### What was done
* A new method SetCellData() was created;

### Test Steps
1. Add an action on OnCellValueChange available on the column
2. Inside this action use the SetCellData client action to change a value of another cell inside this row.
3. Check if the second cell is edited, after the first one is edited. 

### Checklist
* [x] tested locally
* [x] documented the code
* [x] requires changes in OutSystems
* [x] requires new sample page in OutSystems


code/src/GridAPI/Cells.ts
```diff
@@ -40,4 +40,34 @@ namespace GridAPI.Cells {
...
+        gridID: string,
+        rowIndex: number,
+        columnID: string,
+        // eslint-disable-next-line
+        value: any,
+        showDirtyMark = true
+    ): void {
+        if (OSFramework.Helper.IsGridReady(gridID) === false) return;
+        const grid = GridManager.GetGridById(gridID);
+        const column = ColumnManager.GetColumnById(columnID).provider;
```
tmbp95, line 63: We should avoid using the provider here and inside any other API method

code/src/GridAPI/ContextMenu.ts
```diff
@@ -17,8 +17,9 @@ namespace GridAPI.ContextMenu {
     ): string {
         //Try to find in DOM only if not present on Map
         if (lookUpDOM && !_menuItemsToGridId.has(menuItemId)) {
-            const menuOptionElement =
-                OSFramework.Helper.GetElementByUniqueId(menuItemId);
+            const menuOptionElement = OSFramework.Helper.GetElementByUniqueId(
```
tmbp95, line 20: Please run lint here and auto-fix this error

code/src/OSFramework/Interface/IDataSource.ts
```diff
@@ -69,5 +69,10 @@ namespace OSFramework.Grid {
          * @param dataItem
          */
         toOSFormat(dataItem: any): any;
+        /**
+         * Transform value to JSON format
```
tmbp95, line 73: By the name of the method I would not think that the method was going to transform value to JSON format :p

code/src/WijmoProvider/Features/CellData.ts
```diff
@@ -0,0 +1,39 @@
...
+        public build(): void {
+            //
+        }
+
+        public setCellData(
+            rowNumber: number,
+            columnID: string,
+            value: string
+        ): void {
+            const column = GridAPI.ColumnManager.GetColumnById(columnID);
```
tmbp95, line 25: We should avoid doing API calls inside the provider

code/src/WijmoProvider/Helper/DataUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+        if (hasTimeZone) {
+            ticks = value.getTime();
+        } else {
+            ticks = value.getTime() - value.getTimezoneOffset() * 60 * 1000;
+        }
+
+        return ticks;
+    }
+
+    export function TrimSeconds(value: string): string {
```
tmbp95, line 18: What is the difference between this method and the "trimSecondsFromDate" previously defined?

code/src/OSFramework/Interface/IDataSource.ts
bmartinho: Aren't you looking at the description of the next method?

code/src/WijmoProvider/Helper/DataUtils.ts
bmartinho: The TrimSecondsForDate will trim the seconds: 2021-06-25 12:23:34 => 2021-06-25 12:23 -> This is required format for SetCellData from wijmo

The TrimSeconds will clean the seconds: 2021-06-25T12:23:34Z => 2021-06-25T12:23:00Z -> This is required for comparisons in Conditional Format - Will rename this!

code/src/OSFramework/Feature/ICellData.ts
```diff
@@ -10,7 +10,7 @@ namespace OSFramework.Feature {
          */
         setCellData(
             rowNumber: number,
-            columnID: string,
+            columnID: OSFramework.Column.IColumn,
```
tmbp95, line 13: can you please change this to column instead of columnID

bmartinho (approved):
Ready to merge

tmbp95 (approved):

PR #108, opened 2021-06-30
Release 2.3.0

glundgren93:
# Added functionalities
- Addition of Calculated Column.
- Addition of new public client actions for Row and Style.
- Addition of SetCellData public client action.
As suggested by @PedroMSoares.
- Addition of new parameter on MarkChangesAsSaved that allows to clean all dirty marks.
- Addition of a new input parameter on columns that allow setting if the column be hidden in the column picker menu or not.
Kudos to @rmadeiraneto  for providing a solution on github.
- All exported files can now have a custom name.
- Dropdown column can now be bound with another dropdown column.

# Fixed issues
- Fixed bug where Selection API was returning wrong values.
- Fixed bug where grid didn’t render correctly when changing data source.
- Fixed bug where conditional format was not being applied on exported files.
- Fixed bug where adding new row wasn’t working properly.


tmbp95 (approved):

bmartinho (approved):
Ready to go

PR #110, opened 2021-07-02
Fix for issue with getChangesMade client action

bmartinho:
This PR is for fixing 

### What was happening
* There is a bug in the getChangesMade client action that is returning the same row in AddedItems and EditedItems, when an end user adds a new row and edits that same row.

### What was done
* Since we are now using Calculated Columns, the row data item is no longer an object but a Proxy to the data item object. Instead of adding the object in the itemsAdded array, we are now adding the corresponding Proxy object from the data source.

### Test Steps
1. Open the sample
2. Add a new Row
3. Add values into the cells
4. Click outside the row
5. Run GridAPI.GridManager.GetActiveGrid().getChangesMade() and check that there is only information in AddedRows and nothing in EditedRows

### Screenshots
image

### Checklist
* [x] tested locally
* [x] documented the code


glundgren93 (approved):
LGTM

tmbp95 (approved):
Ready to go

PR #115, opened 2021-07-07
RGRIDT-894: Fixing GetRowData

glundgren93:
This PR fixes GetRowData when Grid is Single Entity.

### What was happening
* GetRowData was not working properly on grids with single entity.

### What was done
* Created method that check is grid is single entity or not and returns dataItem from row.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/Rows.ts
+++ b/code/src/WijmoProvider/Features/Rows.ts
@@ -267,6 +267,7 @@ namespace WijmoProvider.Feature {
             ) as CssClassInfo;
         }
 
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         public getRowData(rowNumber: number): any {
             return this._grid.isSingleEntity
                 ? OSFramework.Helper.Flatten(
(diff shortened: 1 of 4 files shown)


JoaoFerreira-FrontEnd (approved):
👍

code/src/WijmoProvider/Features/Rows.ts
```diff
@@ -266,6 +267,16 @@ namespace WijmoProvider.Feature {
             ) as CssClassInfo;
         }
 
+        public getRowData(rowNumber: number): string {
```
tmbp95, line 270: I would make it different. This method might be useful for other use cases and in my opinion, it should not return a stringified JSON but instead the dataItem. Then, in the GridAPI you can use the stringify whenever you need it to be a string.

glundgren93: makes sense.

glundgren93: Changed it

tmbp95 (approved):

tmbp95 (dismissed):

tmbp95 (approved):

JoaoFerreira-FrontEnd (approved):

rugoncalves (approved):

PR #116, opened 2021-07-14
RGRIDT-919: Fixing dropdown on cell value change

glundgren93:
[Sample page](url)

### What was happening
* Whenever we changed values on parent dropdowns sequentially, an error was thrown on console. This was happening because we were returning unformatted data from cell.
dropdownValuechange

### What was done
* Adjusted wijmo.getCellData method to return formatted values.

### Screenshots
(prefer animated gif)

### Checklist
* [x] tested locally
* [ ] documented the code

### DIFF
--- a/code/src/WijmoProvider/Columns/DropdownColumn.ts
+++ b/code/src/WijmoProvider/Columns/DropdownColumn.ts
@@ -33,7 +33,8 @@ namespace WijmoProvider.Column {
             if (oldValue !== newValue) {
                 const currentValue = this.grid.provider.getCellData(
                     rowNumber,
-                    this.provider.index
+                    this.provider.index,
+                    true
                 );
 
                 this.grid.features.dirtyMark.saveOriginalValue(
@@ -46,7 +47,7 @@ namespace WijmoProvider.Column {
                     '',
                     true
                 );
-                this.grid.features.validationMark.validateRow(rowNumber);
+                this.grid.features.validationMark.validateCell(rowNumber, this);
 
                 const column = this.grid.getColumn(columnID);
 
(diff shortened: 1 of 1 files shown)


code/src/WijmoProvider/Columns/DropdownColumn.ts
```diff
@@ -169,7 +179,9 @@ namespace WijmoProvider.Column {
                     this.config.dropdownOptions = values;
                     dataMap.collectionView.sourceCollection = values;
                     dataMap.collectionView.refresh();
-                    if (this.config.parentBinding) {
+
+                    // eslint-disable-next-line no-extra-boolean-cast
+                    if (!!this.config.parentBinding) {
```
JoaoFerreira-FrontEnd, line 184: What does this mean? (!!)

code/src/WijmoProvider/Columns/DropdownColumn.ts
```diff
@@ -116,10 +120,13 @@ namespace WijmoProvider.Column {
                 this.config.dropdownOptions &&
                 this.config.dropdownOptions.length > 0
             ) {
-                this.changeDisplayValues();
-                if (!this._handlerAdded) {
-                    this._parentHandler();
-                    this._handlerAdded = true;
+                // eslint-disable-next-line no-extra-boolean-cast
+                if (!!this.config.parentBinding) {
```
JoaoFerreira-FrontEnd, line 124: What does this mean? (!!)

code/src/WijmoProvider/Columns/DropdownColumn.ts
tmbp95:
```
It basically checks if the variable in question has a value, making a cast to boolean.

- "",  0, null, undefined and NaN will always return False.
- {}, [], "string", 999, new Date() will always return True.
```

I'm not sure about using it here. But anyway, it's checking if the **parentBinding** is neither undefined nor empty.

code/src/WijmoProvider/Columns/DropdownColumn.ts
```diff
@@ -128,9 +135,12 @@ namespace WijmoProvider.Column {
             const dataMap = this.config.dataMap;
             const values = dataMap.collectionView.items;
 
-            const column = this.grid.getColumn(this.config.parentBinding);
+            const parentColumn = this.grid.getColumn(this.config.parentBinding);
```
tmbp95, line 138: What if this.config.parentBinding is undefined, what's the expected result from the getColumn?

glundgren93: Undefined, but I check for it on the if below

tmbp95: Ok, I was concerned about the result of the getColumn when the parentBinding was undefined.

tmbp95 (approved):

code/src/WijmoProvider/Columns/DropdownColumn.ts
bmartinho: Use !! to cast to boolean and check if it is true

bmartinho (approved):
Ready to ship

PR #117, opened 2021-07-14
RGRIDT-923: Improving the way that osRowMetadata is removed

rugoncalves:
This PR is for improving the way metadata is being removed from the lines when they are being returned to the OutSystems side.

### What was happening
* When the action column link was clicked, the osRowMetadata was being removed from the actual line, causing the dirty marks, errors and other css classes to be removed in the line next refresh (i.e. after scroll).
* The metadata information, was not being removed when the API GetChangedLines was called. 

### What was done
* The action column, now performs a clone of the line before removing the metadata and returning it
* The action column, additionally invokes the method `clear` of the `RowMetadata` object that knows how to remove the metadata.
* The data source (AbstractDataSource), is now aware of who the parent grid is, and in the method `_getChangesString` now invokes the `clear` method  of the object `RowMetadata`.

### Test Steps
1. Go to a grid that is editable and has an Action column
2. Make some changes in one of the lines
3. Notice that the dirty marks should appear
4. Click on the action column link
5. Scroll in the data grid until the line disappears
6. Scroll back to the line
7. Notice that the dirty mark is still there
-----
1. Go to a grid that is editable
2. Make some changes in one of the lines
3. Notice that the dirty marks should appear
4. Click to Save Changes (or any other button that invokes the API GetChangedLines
5. Validate that the return value of the GetChangedLines, does not have any property named __osRowMetadata

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/OSFramework/Grid/AbstractDataSource.ts
+++ b/code/src/OSFramework/Grid/AbstractDataSource.ts
@@ -107,6 +107,7 @@ namespace OSFramework.Grid {
         // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
         protected _ds: Array<any>;
         protected _metadata: JSON;
+        protected _parentGrid: IGrid;
 
         constructor() {
             // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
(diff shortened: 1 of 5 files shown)


glundgren93 (approved):

bmartinho (approved):
LGTM

PR #121, opened 2021-07-20
RGRIDT-869: Improving Cell Style metadata

glundgren93:
### What was happening
* Cell style metadata only was able to have one class, since it was a string.

### What was done
* Changed cell style metadata to have a string array, so a cell can have multiple cells. Adjusted conditional format to support this as well.

### Test Steps
1. Press Show row cell metadata
2. Sample_Product.Price should only have background-yellow
3. Press Set Price Cell class
4. Sample_Product.Price should have background-yellow and background-pink

### Screenshots
(prefer animated gif)

### Checklist
* [x] tested locally
* [x] documented the code


tmbp95 (approved):
Looking good! 🚀

JoaoFerreira-FrontEnd (approved):
Nice Job!

PR #123, opened 2021-07-30
RGRIDT-933: Server-side Pagination

glundgren93:
This PR contains the changes necessary for server-side pagination.

### What was done
* Create the ServerSidePagination input parameter
* Hide grid pagination when ServerSidePagination is true
* Show only Filter by Condition when ServerSidePagination is true
* Update API Activate filter to show only Filter by Condition when ServerSidePagination is true
* Create Client action to set the first value of grid auto numbering
* Create the OnSortChange event

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/GridAPI/Rows.ts
+++ b/code/src/GridAPI/Rows.ts
@@ -119,7 +119,10 @@ namespace GridAPI.Rows {
      * @param {string} gridID ID of the Grid where the change will occur.
      * @param {number} startIndex New row start index.
      */
-    export function UpdateRowNumber(gridID: string, startIndex: number): void {
+    export function UpdateStartingRowHeader(
+        gridID: string,
+        startIndex: number
+    ): void {
         const grid = GridManager.GetGridById(gridID);
 
         if (grid !== undefined) {
--- a/code/src/GridAPI/Rows.ts
+++ b/code/src/GridAPI/Rows.ts
@@ -111,4 +111,19 @@ namespace GridAPI.Rows {
 
         return output;
     }
+
+    /**
+     * Function that will set start index of row.
+     *
+     * @export
+     * @param {string} gridID ID of the Grid where the change will occur.
+     * @param {number} startIndex New row start index.
+     */
+    export function UpdateRowNumber(gridID: string, startIndex: number): void {
+        const grid = GridManager.GetGridById(gridID);
+
+        if (grid !== undefined) {
+            grid.features.autoRowNumber.setStartIndex(startIndex);
+        }
+    }
 }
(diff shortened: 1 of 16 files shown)


code/src/GridAPI/Rows.ts
```diff
@@ -111,4 +111,19 @@ namespace GridAPI.Rows {
...
     }
+
+    /**
+     * Function that will set start index of row.
+     *
+     * @export
+     * @param {string} gridID ID of the Grid where the change will occur.
+     * @param {number} startIndex New row start index.
+     */
+    export function UpdateRowNumber(gridID: string, startIndex: number): void {
```
tmbp95, line 122: UpdateRowNumber name might be confusing as it sounds like you will be able to update A row number instead of changing the startIndex. (Not sure if I'm confused with this or this is confusing me 😅)

code/src/WijmoProvider/Features/ColumnFilter.ts
```diff
@@ -67,11 +68,14 @@ namespace WijmoProvider.Feature {
...
+
         public activate(columID: string): void {
             const column = GridAPI.ColumnManager.GetColumnById(columID);
 
-            this._filter.getColumnFilter(column.provider).filterType =
-                wijmo.grid.filter.FilterType.Both;
+            this._filter.getColumnFilter(column.provider).filterType = this
+                ._grid.config.serverSidePagination
+                ? wijmo.grid.filter.FilterType.Condition
+                : wijmo.grid.filter.FilterType.Both;
```
tmbp95, line 78: It looks good! 🔝
If we end up using this condition multiple times, please add it into a new getter or a method that will return FilterType.Condition or FilterType.Both depending on the serverSidePagination or any other condition that we might end up adding into the formula. (I'm seeing this again inside the setState method)

tmbp95 (approved):

code/src/GridAPI/Rows.ts
JoaoFerreira-FrontEnd: Could be UpdateStartIndex?

JoaoFerreira-FrontEnd (approved):

rugoncalves (approved):
It would be great to have some more comments in the code

code/src/WijmoProvider/Features/ColumnFilter.ts
```diff
@@ -67,11 +68,18 @@ namespace WijmoProvider.Feature {
                 ).length > 0
             );
         }
+
+        public get filterType(): wijmo.grid.filter.FilterType {
```
tmbp95, line 72: 🔝 🔝 🔝

tmbp95 (dismissed):

tmbp95 (approved):

PR #129, opened 2021-08-20
RGRIDT-973: Set rows valid after deletion

glundgren93:
If we had one invalid line and removed it from the grid, the line would appear on InvalidLines and RemovedLines on GetChanges method. This PR addresses this issue, by setting all rows valid after they are removed.

### What was happening
* If we had one invalid line and removed it from the grid, the line would appear on InvalidLines and RemovedLines on GetChanges method. 

### What was done
* Created new method to set row status and turn rows valid after deletion

### Test Steps
1. Turn a row invalid and then remove it.
2. Call GetChanges API.
3. InvalidLines JSON should be empty

### Checklist
* [x] tested locally
* [ ] documented the code

### DIFF
--- a/code/src/OSFramework/Feature/IValidationMark.ts
+++ b/code/src/OSFramework/Feature/IValidationMark.ts
@@ -7,13 +7,13 @@ namespace OSFramework.Feature {
         errorMessage(rowNumber: number, binding: string): string;
         isInvalid(rowNumber: number, binding: string): boolean;
         isInvalidRow(row: any): boolean;
-        setRowStatus(rowNumber: number, isValid: boolean): void;
-        setStatus(
+        setCellStatus(
             rowNumber: number,
             columnID: string,
             isValid: boolean,
             errorMessage: string
         ): void;
+        setRowStatus(rowNumber: number, isValid: boolean): void;
         validateCell(
             rowNumber: number,
             column: OSFramework.Column.IColumn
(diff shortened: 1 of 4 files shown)


code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -399,6 +400,11 @@ namespace WijmoProvider.Feature {
             );
         }
 
+        public setRowStatus(rowNumber: number, isValid: boolean): void {
+            // set invalidRows with row number and flag that checks if status isValid and if there are invalid values on metadata
+            this._setInvalidRows(rowNumber, isValid);
```
tmbp95, line 405: the method this._setInvalidRows should be renamed as well 🤔

code/src/OSFramework/Feature/IValidationMark.ts
```diff
@@ -7,6 +7,7 @@ namespace OSFramework.Feature {
         errorMessage(rowNumber: number, binding: string): string;
         isInvalid(rowNumber: number, binding: string): boolean;
         isInvalidRow(row: any): boolean;
+        setRowStatus(rowNumber: number, isValid: boolean): void;
         setStatus(
```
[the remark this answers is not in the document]
glundgren93: fixed

code/src/WijmoProvider/Features/ValidationMark.ts
glundgren93: fixed

tmbp95 (dismissed):
nice one!

rugoncalves (dismissed):

PR #131, opened 2021-08-26
RGRIDT-934: Server side sync

glundgren93:
This PR delivers the necessary changes for server-side synchronization on grid.

### What was done
* RGRIDT-934: Added new KeyBinding attribute.
* RGRIDT-967: Added new UpdateAddedLineKey API method.
* RGRIDT-968: Added new GetRowNumberByKey API method.
* RGRIDT-990: Added new SetValidationStatusByKey API method.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/OSFramework/Grid/AbstractDataSource.ts
+++ b/code/src/OSFramework/Grid/AbstractDataSource.ts
@@ -212,8 +212,10 @@ namespace OSFramework.Grid {
         }
 
         public getRowNumberByKey(key: string): number {
-            return this._ds.findIndex(
-                (item) => _.get(item, this.parentGrid.config.keyBinding) === key
+            return this.parentGrid.provider.rows.findIndex(
+                (item) =>
+                    _.get(item.dataItem, this.parentGrid.config.keyBinding) ===
+                    key
             );
         }
 
@@ -260,7 +262,7 @@ namespace OSFramework.Grid {
             return value;
         }
 
-        public updateAddedLineKey(
+        public updateAddedRowKey(
             currentRowId: string,
             newKey: string
         ): boolean {
--- a/code/src/OSFramework/Grid/AbstractDataSource.ts
+++ b/code/src/OSFramework/Grid/AbstractDataSource.ts
@@ -27,7 +27,8 @@ namespace OSFramework.Grid {
     ): any {
         //regex expressions for date and datetime should be described here
         const regex = {
-            datetime: /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2}(?:\.\d*)?)Z?$/, //yyyy-MM-ddThh:mm:ssZ
+            datetime:
+                /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2}(?:\.\d*)?)Z?$/, //yyyy-MM-ddThh:mm:ssZ
             date: /^(\d{4})-(\d{2})-(\d{2})$/ //yyyy-MM-dd
         };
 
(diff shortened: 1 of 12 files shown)


code/src/OSFramework/Grid/AbstractDataSource.ts
```diff
@@ -116,6 +118,22 @@ namespace OSFramework.Grid {
             this._isSingleEntity = false;
         }
 
+        private _getRowByKey(key: string) {
+            return this._ds.find((item) => {
```
tmbp95, line 122: Can we get the row by getting the rownumber (with the getRowNumberByKey ) and then just use the rowNumber to get the row?

code/src/OSFramework/Interface/IDataSource.ts
```diff
@@ -83,5 +85,7 @@ namespace OSFramework.Grid {
          * @param value
          */
         trimSecondsFromDate(value: string): string;
+
+        updateAddedRowKey(currentRowId: string, newKey: string): boolean;
```
tmbp95, line 89: Can you please add a comment here so that we have a little help while using the method in the future?

code/src/OSFramework/Interface/IDataSource.ts
```diff
@@ -59,6 +59,8 @@ namespace OSFramework.Grid {
          */
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         getProviderDataSource(): any;
+
+        getRowNumberByKey(key: string): number;
```
tmbp95, line 63: Can you please add a comment here so that we have a little help while using the method in the future?

code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -112,6 +112,32 @@ namespace WijmoProvider.Feature {
             return false;
         }
 
+        /**
+         * Checks if a specific row is invalid by checking if it contains any invalid cells.
+         * @param rowKey Key of the row to check if a row has invalid cells
+         * @returns Boolean that indicates if the row is invalid. Returns True if invalid. False, otherwise.
+         */
+        private _isInvalidRowByRowKey(rowKey: string): boolean {
```
tmbp95, line 120: can we optimize this method as is looks a lot similar to the older "_isInvalidRowByRowNumber"

tmbp95 (commented):
Is there a way to reuse the older methods that use the RowByNumber? We are repeating the code all over the files. I understand that some methods are relative to the RowByKey and the other by Number, but can't we use the same method to do the same thing? 
If the argument is a string, then it's by key, if the argument is a number, then it's by number..
This is just a question, open to be discussed!

glundgren93:
> Is there a way to reuse the older methods that use the RowByNumber? We are repeating the code all over the files. I understand that some methods are relative to the RowByKey and the other by Number, but can't we use the same method to do the same thing?
> If the argument is a string, then it's by key, if the argument is a number, then it's by number..
> This is just a question, open to be discussed!

That would be possible if we didn't have the key as a number when we add new rows. I created new methods that practically do the same thing in order to help us in the future if we need to change some behavior on a specific use case. If you think it's better to have one abstract method that handles all this, I can change the code.

tmbp95:
> > Is there a way to reuse the older methods that use the RowByNumber? We are repeating the code all over the files. I understand that some methods are relative to the RowByKey and the other by Number, but can't we use the same method to do the same thing?
> > If the argument is a string, then it's by key, if the argument is a number, then it's by number..
> > This is just a question, open to be discussed!
> 
> That would be possible if we didn't have the key as a number when we add new rows. I created new methods that practically do the same thing in order to help us in the future if we need to change some behavior on a specific use case. If you think it's better to have one abstract method that handles all this, I can change the code.

I think @rugoncalves can have a better opinion here. I'm not sure about the best option, as I said 😅

code/src/WijmoProvider/Features/ValidationMark.ts
glundgren93: Please check now.

rugoncalves (approved):
More comments in the code, would be greatly welcome.

JoaoFerreira-FrontEnd (approved):

PR #133, opened 2021-08-31
RGRIDT-994: Adding has thousand separator property

glundgren93:
This PR adds hasThousandSeparator property to number columns

### What was happening
* It was not possible to have a column with year, because of the thousand separators. The year 2021 would become 2,021.

### What was done
* Added a property that indicates whether or not the number column should have a thousand separators.

### Checklist
* [x] tested locally
* [x] documented the code


rugoncalves (approved):
Looking good!

JoaoFerreira-FrontEnd (approved):

PR #134, opened 2021-09-01
RGRIDT-899: Adding format to date/dateTime columns

glundgren93:
This PR adds format to Date and DateTime columns.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Columns/DateColumn.ts
+++ b/code/src/WijmoProvider/Columns/DateColumn.ts
@@ -22,7 +22,8 @@ namespace WijmoProvider.Column {
                 columnID,
                 new OSFramework.Configuration.Column.ColumnConfig(configs),
                 new OSFramework.Configuration.Column.EditorConfigDate(
-                    editorConfig
+                    editorConfig,
+                    false
                 )
             );
             this._columnEvents =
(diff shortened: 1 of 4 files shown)


code/src/WijmoProvider/Columns/DateColumn.ts
```diff
@@ -49,7 +49,7 @@ namespace WijmoProvider.Column {
 
         public build(): void {
             //Setting date format
-            this.config.format = GridAPI.dateFormat;
+            this.config.format = this.editorConfig.format || GridAPI.dateFormat;
```
tmbp95, line 52: Is this the only way to get the dateFormat? We should avoid accessing the GridAPI on the wijmoProvider

code/src/WijmoProvider/Columns/DateTimeColumn.ts
```diff
@@ -48,7 +48,9 @@ namespace WijmoProvider.Column {
 
         public build(): void {
             this.editorConfig.timeFormat = 'HH:mm';
-            this.config.format = `${GridAPI.dateFormat} ${this.editorConfig.timeFormat}`;
+            this.config.format =
+                this.editorConfig.format ||
+                `${GridAPI.dateFormat} ${this.editorConfig.timeFormat}`;
```
tmbp95, line 53: Is this the only way to get the dateFormat? We should avoid accessing the GridAPI on the wijmoProvider

code/src/WijmoProvider/Columns/DateColumn.ts
rugoncalves: I agree with Tiago. Also these changes are they anyhow typed in OS side? as in, can we have them typed somehow (e.g. enum)?

code/src/WijmoProvider/Columns/DateTimeColumn.ts
rugoncalves: Agree, we should preferably always use the config, now that we have it. Meaning that when we're loading the configs we should set the value if they come empty, and from them on, the configs are "master".

rugoncalves (changes requested):

rugoncalves (approved):

JoaoFerreira-FrontEnd (approved):

PR #135, opened 2021-09-07
Adding method GetCurrentPage to the API

rugoncalves:
This PR is for adding the GetCurrentPage API, to return the information about which page we currently are.

### What was happening
* There was no **public api**, to return the current page number
* This was requested by the user: [Alfio Esposito](https://www.outsystems.com/profile/ddjduo8qiw/), in the [forum](https://www.outsystems.com/forums/discussion/72996/activate-a-page-in-the-grid-dynamically/#Post298621).

### What was done
* Added a public API that exposes the _pageIndex_ getter of the **pagination feature**.
* No other change was done.

### Checklist
* [ ] tested locally
* [x] documented the code


code/src/GridAPI/Pagination.ts
```diff
@@ -60,6 +60,31 @@ namespace GridAPI.Pagination {
...
+     * Gets the current page Index of the DataGrid.
+     *
+     * @export
+     * @param {string} gridID Id of the Grid from which to obtain the pagination Index
+     * @return {*}  {number} Index of the current page, 0 based. Return -1 if the Grid is not yet initialized.
+     */
+    export function GetCurrentPage(gridID: string): number {
+        PerformanceAPI.SetMark('Pagination.GetCurrentPage');
+
+        if (!OSFramework.Helper.IsGridReady(gridID)) return -1;
```
tmbp95, line 73: Should we throw an error here?

tmbp95 (approved):

glundgren93 (approved):

code/src/GridAPI/Pagination.ts
rugoncalves: Uhm, I think that it would be too rough to throw an error...

PR #136, opened 2021-09-09
RGRIDT-900: Creating Filter by value and filter by condition actions

glundgren93:
This PR delivers filter by value and by condition functions.

### What was done
* Created API methods to filter by value and condition.
* Created methods on IColumnFilter and implemented them on ColumnFilter.

### Checklist
* [x] tested locally
* [x] documented the code


code/src/WijmoProvider/Features/ColumnFilter.ts
```diff
@@ -108,6 +108,62 @@ namespace WijmoProvider.Feature {
             this.setState(this._enabled);
         }
 
+        public byCondition(
+            columnId: string,
+            values: OSFramework.OSStructure.FilterCondition[]
+        ): void {
+            const column = this._grid.getColumn(columnId);
+            if (column) {
+                const columnFilter = this._filter.getColumnFilter(
```
tmbp95, line 117: This line will work even if the column has the filter disabled, right?

tmbp95 (dismissed):

code/src/WijmoProvider/Features/ColumnFilter.ts
```diff
@@ -108,6 +108,62 @@ namespace WijmoProvider.Feature {
...
+                    column.config.binding
+                ).conditionFilter;
+
+                if (values.length > 0) {
+                    const condition1 = values[0];
+                    const condition2 = values[1];
+
+                    columnFilter.condition1.value =
+                        column.columnType === OSFramework.Enum.ColumnType.Number
+                            ? parseInt(condition1.value)
```
bmartinho, line 127: If column type is Number, the code is doing parseInt, shouldn't be for parseFloat? What about for Currency, Date and DateTime?

code/src/WijmoProvider/Features/ColumnFilter.ts
glundgren93: Yes!

code/src/WijmoProvider/Features/ColumnFilter.ts
glundgren93: You're right! Changed to parseFloat. Currency, date and datetime column expect a string.

rugoncalves (dismissed):

tmbp95 (dismissed):

PR #154, opened 2021-11-02
RGRIDT-1051: Fixing issue when adding rows and removing wrong use of API

glundgren93:
This PR is for a bug fix reported on the [forum](https://www.outsystems.com/forums/discussion/74212/adding-row-when-using-dependant-dropdowns-gives-error/) by [Bård Indredavik](https://www.outsystems.com/profile/p6p2u4tgiv/).

### What was happening
* Whenever we tried to add rows on the grid with dependent dropdowns, an error was thrown. This was happening because the grid behaved as if the filters were applied.

### What was done
* Added an extra validation on the filter test.
* Removed wrong use of GridAPI on wijmoProvider.

### Test Steps
1. Open the context menu and add row.
2. Row should be added.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/ColumnFilter.ts
+++ b/code/src/WijmoProvider/Features/ColumnFilter.ts
@@ -62,9 +62,11 @@ namespace WijmoProvider.Feature {
         }
 
         public get isGridFiltered(): boolean {
+            // when filter is active, the filterDefinition object usually has filterType different than 0
+            // and it has a type. we should check for both.
             return (
                 JSON.parse(this._filter.filterDefinition).filters.filter(
-                    (x) => x.filterType !== 0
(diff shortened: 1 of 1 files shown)


tmbp95 (approved):

rugoncalves (changes requested):
Well done! Liked specially the fix that you did to remove the calls do the grid API.
Just a minor thing to change, can you give a different name to x? what is x?
I know that it was already like that, but since you're changing that line, let's improve it!

rugoncalves (approved):
Approved! :)

tmbp95 (approved):

JoaoFerreira-FrontEnd (approved):

PR #156, opened 2021-11-03
RGRIDT-1039: Add/remove rows according to selection/checkbox

glundgren93:
This PR adjusts the necessary methods to ensure row addition/removal respects cell selection and row checkbox selection.

### What was happening
* With the newly added feature of checkbox, whenever we had a row checked and a cell on the same row selected, the grid considered there were two selections, and it would add/remove two rows.

### What was done
* Adjusted getProviderAllSelections method to remove intersection between checkedRows and cellSelection from same rows.
* On a previous task, the grid's selection mode was set to CellRange when we had Checkbox. This mode didn't allow us to select multiple cells with CTRL. Switched back to MultiRange. 

### Test Steps
1. Check a row.
2. Select a cell on the same row.
3. Open the context menu and add a row.

Expected result: Only one row should've been added.

1. Check a row.
2. Select a cell from another row.
3. Open the context menu and add a row.

Expected result: Two rows should've been added.

1. Check a row.
2. Change pages and check another row
3. Open the context menu and add a row.

Expected result: Only one row should've been added.

Note: If a row is checked and a cell is selected, the grid will always consider two selections and will add two rows.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/Selection.ts
+++ b/code/src/WijmoProvider/Features/Selection.ts
@@ -318,15 +318,30 @@ namespace WijmoProvider.Feature {
                 ...this._grid.provider.selectedRanges.filter((p) => p.isValid)
             );
             // }
-            return this._getCheckedRows()
+
+            // create checkedRows cell range.
+            let checkedRowsRange = this._getCheckedRows()
                 .map((p) => new wijmo.grid.CellRange(p, 0, p, maxCol))
                 .filter((p) => {
                     for (let i = 0; i < ranges.length; i++) {
                         if (ranges[i].contains(p)) return false;
                     }
                     return true;
-                })
-                .concat(ranges);
+                });
+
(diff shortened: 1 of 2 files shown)


code/src/WijmoProvider/Features/Selection.ts
```diff
@@ -318,15 +318,30 @@ namespace WijmoProvider.Feature {
                 ...this._grid.provider.selectedRanges.filter((p) => p.isValid)
             );
             // }
-            return this._getCheckedRows()
+
+            // create checkedRows cell range.
+            let checkedRowsRange = this._getCheckedRows()
```
tmbp95, line 323: The name could be different as checkedRowsRange sounds like you are only going to have the rows that have been checked ☑️.

tmbp95 (approved):

JoaoFerreira-FrontEnd (approved):

PR #161, opened 2021-11-10
RGRIDT-1062: Equalize selection was working when we had checked rows.

glundgren93:
Limiting equalizeSelection to work only if grid doesn't have checked rows

### What was happening
* Since we decided to follow wijmo's behavior (only copy checked content) when dealing with Copy, checkbox, and cell selection, whenever we tried to copy data from a grid that had checked rows and cell selection, the grid was equalizing the selection, but when we pasted the content, only the checked row was there. This was misleading, as it seemed everything would be copied.
equalize

### What was done
* Added an extra condition on equalizeSelection method to check if grid has checked rows. If it does, we don't want to equalize.

### Test Steps
1. On Grid's without checkbox, select a cell from one column and then select a cell from a different row and column.
2. Press control C
3. Grid should have equalized the selection

1. On Grid's with checkbox, select a cell from one column and then select a cell from a different row and column.
2. Press control C
3. Grid should have equalized the selection

1. On Grid's with checkbox, select a cell from one column and then select a different row using the checkbox.
2. Press control C
3. Grid should have not equalized the selection

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/Selection.ts
+++ b/code/src/WijmoProvider/Features/Selection.ts
@@ -146,7 +146,7 @@ namespace WijmoProvider.Feature {
         }
 
         public equalizeSelection(): OSFramework.OSStructure.CellRange[] {
-            //This method just makes sense for MultiRange and for grid's without checked rows
+            //This method just makes sense for MultiRange or for grid's without checked rows
             if (
                 this._grid.provider.selectionMode !==
                     wijmo.grid.SelectionMode.MultiRange ||
--- a/code/src/WijmoProvider/Features/Selection.ts
+++ b/code/src/WijmoProvider/Features/Selection.ts
@@ -146,10 +146,11 @@ namespace WijmoProvider.Feature {
         }
 
         public equalizeSelection(): OSFramework.OSStructure.CellRange[] {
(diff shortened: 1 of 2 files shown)


code/src/WijmoProvider/Features/Selection.ts
```diff
@@ -146,10 +146,11 @@ namespace WijmoProvider.Feature {
         }
 
         public equalizeSelection(): OSFramework.OSStructure.CellRange[] {
-            //This method just makes sense for MultiRange
+            //This method just makes sense for MultiRange and for grid's without checked rows
             if (
                 this._grid.provider.selectionMode !==
-                wijmo.grid.SelectionMode.MultiRange
+                    wijmo.grid.SelectionMode.MultiRange ||
```
tmbp95, line 152: just to make the point that you wrote "and for grid's without checked rows" and then used a || here

glundgren93: changed it to "or"

tmbp95 (approved):

rugoncalves (approved):
Heads-up, this behavior should be documented in the Sample/Documentation.

PR #166, opened 2021-11-30
RGRIDT-1063: Refactoring conditional format

glundgren93:
This PR removes some abstraction on conditional format.

### What was happening
* Conditional format was being applied on a single method that sealed formatting for all column types. This made things harder to maintain.

### What was done
* Created a new method to evaluate text cells.


tmbp95 (approved):
🚀🚀🚀

JoaoFerreira-FrontEnd (approved):

PR #171, opened 2021-12-09
RGRIDT-1094: Adding success code to Add/Remove row

glundgren93:
This PR is for a [issue ]([Data Grid Reactive] AddNewRows and RemoveSelectedRows actions always returning false in output parameter)reported on forum.

### What was happening
* Add/Remove rows were always returning false because on our client actions our success output parameter is set as 
image

### What was done
* Added success code.

### Test Steps
1. Add row.
2. Success message should be shown on green alert.


bmartinho (approved):
Ready to go

JoaoFerreira-FrontEnd (approved):

PR #172, opened 2021-12-14
RGRIDT-1099: GetChangedLines throw errors on empty Grid that has multiple entities

glundgren93:
This PR fixes IsSingleEntity returning wrong value on empty grids with multiple entities.

### What was happening
* Empty grids with multiple entities were considered to be single entity.

### What was done
* Check metadata keys to determine if grid is single entity or not.

### DIFF
--- a/code/src/OSFramework/Grid/AbstractDataSource.ts
+++ b/code/src/OSFramework/Grid/AbstractDataSource.ts
@@ -247,9 +247,7 @@ namespace OSFramework.Grid {
             this._metadata = dataJson.metadata;
 
             this._isSingleEntity =
-                Object.keys(
-                    this._metadata || dataJson[0] || dataJson.data[0] || {}
-                ).length <= 1;
+                Object.keys(this._metadata || dataJson[0] || {}).length <= 1;
 
             if (this.hasMetadata) {
                 this._ds = [...dataJson.data];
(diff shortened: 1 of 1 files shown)


JoaoFerreira-FrontEnd (dismissed):

code/src/OSFramework/Grid/AbstractDataSource.ts
```diff
@@ -247,7 +247,9 @@ namespace OSFramework.Grid {
             this._metadata = dataJson.metadata;
 
             this._isSingleEntity =
-                Object.keys(dataJson[0] || dataJson.data[0] || {}).length <= 1;
+                Object.keys(
+                    this._metadata || dataJson[0] || dataJson.data[0] || {}
```
rugoncalves, line 251: @glundgren93 you should remove this part `dataJson.data[0]` because:
1. it's not doing anything :)
2. if the developer does not use our action and passes an {}, it will throw an exception...

rugoncalves (changes requested):

code/src/OSFramework/Grid/AbstractDataSource.ts
glundgren93: fixed

rugoncalves (approved):

JoaoFerreira-FrontEnd (approved):

PR #173, opened 2021-12-15
RGRIDT-757: Fixing undo/redo on delete row

glundgren93:
This PR fixes a bug where an error was thrown when we used undo/redo after deleting rows.

### What was happening
* When we used undo/redo after deleting rows, the Validation mark feature was being triggered with the row object. This lead to multiple errors.

### What was done
* Added a condition on Validation mark undo/redo handlers to check whether or not the old state was an object. 

### Test Steps
1. Select rows
2. Press remove select rows
3. Press ctrl z
4. Press ctrl y

Expected result: Rows should have been deleted and no error should be thrown on the console.

### DIFF
--- a/code/src/WijmoProvider/Features/ValidationMark.ts
+++ b/code/src/WijmoProvider/Features/ValidationMark.ts
@@ -157,16 +157,20 @@ namespace WijmoProvider.Feature {
             if (action.dataItem === undefined) return;
 
             const binding = this._grid.provider.getColumn(action.col).binding;
-            const oldValue = this._grid.features.dirtyMark.getOldValue(
-                action.row,
-                binding
-            );
-            this._triggerEventsFromColumn(
-                action.row,
-                binding,
-                oldValue,
-                action._newState
-            );
+
+            // we don't want to redo on GridRemoveRowAction
+            if (typeof action._oldState !== 'object') {
+                const oldValue = this._grid.features.dirtyMark.getOldValue(
+                    action.row,
+                    binding
+                );
+                this._triggerEventsFromColumn(
+                    action.row,
+                    binding,
+                    oldValue,
+                    action._newState
+                );
+            }
         }
 
         /**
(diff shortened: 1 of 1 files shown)


code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -157,16 +157,20 @@ namespace WijmoProvider.Feature {
             if (action.dataItem === undefined) return;
 
             const binding = this._grid.provider.getColumn(action.col).binding;
```
rugoncalves, line 159: This `const binding` can be defined inside the `if (typeof action._oldState !== 'object')`

code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -157,16 +157,20 @@ namespace WijmoProvider.Feature {
...
-            );
-            this._triggerEventsFromColumn(
-                action.row,
-                binding,
-                oldValue,
-                action._newState
-            );
+
+            // we don't want to redo on GridRemoveRowAction
+            if (typeof action._oldState !== 'object') {
```
tmbp95, line 162: Following @rugoncalves suggestion, this one is mine :p

```
private _redoingActionHandler(
    s: wijmo.undo.UndoStack,
    e: wijmo.undo.UndoActionEventArgs
): void {
    const action: any = e.action;
    
    // we only want to redo on GridEditAction
    // we don't want to redo on GridRemoveRowAction
    if (action.dataItem !== undefined && typeof action._oldState !== 'object') {
        const binding = this._grid.provider.getColumn(action.col).binding;
        const oldValue = this._grid.features.dirtyMark.getOldValue(
            action.row,
            binding
        );
        this._triggerEventsFromColumn(
            action.row,
            binding,
            oldValue,
            action._newState
        );
    }
}
```

tmbp95 (dismissed):

code/src/WijmoProvider/Features/ValidationMark.ts
glundgren93: fixed

code/src/WijmoProvider/Features/ValidationMark.ts
glundgren93: fixed

tmbp95 (approved):

JoaoFerreira-FrontEnd (approved):

PR #177, opened 2021-12-22
RGRIDT-1115: Fixing is required on date columns

glundgren93:
This PR fixes a bug when deleting a is required date column.

### What was happening
* There was a [bug ](https://www.outsystems.com/forums/discussion/75258/data-grid-reactive-console-error-when-isrequired-property-for-date-datetime-c/) reported on forum. Whenever we have isRequired on Date/Datetime columns, cell deletion would throw an error.

### What was done
* Adjusted isDirtyMark logic

### Test Steps
1.  To focus/select the cell, click on any Date  column which has value.
2. Press Delete key from keyboard.


bmartinho (approved):
Ready to go!

tmbp95 (approved):

rugoncalves (approved):

JoaoFerreira-FrontEnd (approved):

PR #178, opened 2021-12-23
RGRIDT-1097: Adding SetColumnFilterOptions

glundgren93:
This PR adds a client action that will allow grids with server-side pagination to filter text/dropdown columns by value. The client action will populate the filter list with all possible values.

### Test Steps
1. Filter text column by value.
2. All unique values should be displayed.
3. Select a value.

Expected: Filter should work

1. Filter dropdown column by value.
2. All unique values should be displayed.
3. Select a value.

Expected: Filter should work

### DIFF
--- a/code/src/GridAPI/Filter.ts
+++ b/code/src/GridAPI/Filter.ts
@@ -178,4 +178,39 @@ namespace GridAPI.Filter {
             'Filter.ByValue-end'
         );
     }
+
+    /**
+     * Function that sets column filter options
+     *
+     * @export
+     * @param {string} gridID ID of the Grid block.
+     * @param {string} columnID ID of the column block where the filter options will be set.
+     * @param {string} options  Values that will be used on filter by value list.
+     * @param {number} maxVisibleOptions Maximum number of elements on the filter list of display values.
+     * @returns {*}  {void}
+     */
+    export function SetColumnFilterOptions(
+        gridID: string,
+        columnID: string,
+        options: string,
+        maxVisibleOptions?: number
+    ): void {
+        PerformanceAPI.SetMark('Filter.SetColumnFilterOptions');
+
+        if (!OSFramework.Helper.IsGridReady(gridID)) return;
+        const grid = GridManager.GetGridById(gridID);
+
+        grid.features.filter.setColumnFilterOptions(
+            columnID,
+            JSON.parse(options),
+            maxVisibleOptions
+        );
+
+        PerformanceAPI.SetMark('Filter.SetColumnFilterOptions-end');
+        PerformanceAPI.GetMeasure(
+            '@datagrid-Filter.SetColumnFilterOptions',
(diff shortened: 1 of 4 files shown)


code/src/GridAPI/Filter.ts
```diff
@@ -178,4 +178,39 @@ namespace GridAPI.Filter {
...
+     * @param {string} columnID ID of the column block where the filter options will be set.
+     * @param {string} options  Values that will be used on filter by value list.
+     * @param {number} maxVisibleOptions Maximum number of elements on the filter list of display values.
+     * @returns {*}  {void}
+     */
+    export function SetColumnFilterOptions(
+        gridID: string,
+        columnID: string,
+        options: string,
+        maxVisibleOptions?: number
```
tmbp95, line 196: since this value comes from OutSystems it will be 0 or > 0 right?

code/src/WijmoProvider/Features/ColumnFilter.ts
```diff
@@ -200,6 +198,35 @@ namespace WijmoProvider.Feature {
...
+                column.columnType === OSFramework.Enum.ColumnType.Text ||
+                column.columnType === OSFramework.Enum.ColumnType.Dropdown
+            ) {
+                this._filter.getColumnFilter(
+                    column.provider
+                ).valueFilter.uniqueValues = Array.from(
+                    new Set<string>(options) // we only want unique values
+                );
+
+                if (maxVisibleOptions)
```
tmbp95, line 219: Check my last comment, should we confirm if the value is > 0 here instead?

JoaoFerreira-FrontEnd (approved):

code/src/WijmoProvider/Features/ColumnFilter.ts
glundgren93: Even if its 0, we won't do anything, because it doesn't make sense to show a list with no values, does it?

tmbp95: Ok, so should we have a better and explicit condition here? Or a comment?

JoaoFerreira-FrontEnd: I think we should ignore the 0 since is the empty value for an integer in the SS.

JoaoFerreira-FrontEnd: but we can check if its greater than 0, to make sure we don't config with awkward values

glundgren93: 0 is considered falsy, but ok.

glundgren93: done

tmbp95 (approved):

JoaoFerreira-FrontEnd (approved):

PR #182, opened 2022-01-05
RGRIDT-1122: OnCellChange is sync when column is sorted

glundgren93:
This PR adds a new way to handle onCellChange event when a column is sorted.

### What was happening
* Whenever we had a column sorted and did some validation on the cellChange event, the validation mark was misplaced. This happened because we handled the callback async, which means that our validation happened after Wijmo's sort event.

### What was done
* Created a new method to see if the column is sorted.
* Check if the current cell's column being changed is sorted, if it is we handle the callback in a sync way, otherwise we handle it async (default way)

### Test Steps
1. Sort the Price column

2. Change the value (must be inferior to 100)

3. The reorder is made

Expected: The error is being presented on the line where our new value was reordered

### DIFF
--- a/code/src/OSFramework/Helper/AsyncInvocation.ts
+++ b/code/src/OSFramework/Helper/AsyncInvocation.ts
@@ -4,4 +4,8 @@ namespace OSFramework.Helper {
     export function AsyncInvocation(callback: Callbacks.Generic, ...args: any) {
         if (callback) setTimeout(() => callback(...args), 0);
     }
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
+    export function SyncInvocation(callback: Callbacks.Generic, ...args: any) {
+        if (callback) callback(...args);
+    }
 }
--- a/code/src/WijmoProvider/Features/ColumnSort.ts
+++ b/code/src/WijmoProvider/Features/ColumnSort.ts
@@ -178,6 +178,18 @@ namespace WijmoProvider.Feature {
             );
         }
 
+        public isColumnSorted(columnID: string): boolean {
+            const column = this._grid.getColumn(columnID);
+
+            if (!column) return false;
+
+            return (
+                this._grid.provider.itemsSource.sortDescriptions.find(
+                    (col) => col.property === column.config.binding
+                ) !== undefined
+            );
+        }
+
         public setState(value: boolean): void {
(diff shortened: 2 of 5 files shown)


code/src/WijmoProvider/Features/ColumnSort.ts
```diff
@@ -178,6 +178,18 @@ namespace WijmoProvider.Feature {
             );
         }
 
+        public isColumnSorted(columnID: string): boolean {
+            const column = this._grid.getColumn(columnID);
+
+            if (!column) return false;
+
+            return (
+                this._grid.provider.itemsSource.sortDescriptions.find(
```
tmbp95, line 187: Maybe we should have a property on the column that should get this value (it would be easier). The way to do this is by adding this boolean value every time the OnSort event gets triggered per column.

tmbp95 (approved):
Nice catch!

code/src/WijmoProvider/Features/ColumnSort.ts
rugoncalves: Agree! @glundgren93 ☝️

code/src/OSFramework/Helper/AsyncInvocation.ts
```diff
@@ -4,4 +4,8 @@ namespace OSFramework.Helper {
     export function AsyncInvocation(callback: Callbacks.Generic, ...args: any) {
         if (callback) setTimeout(() => callback(...args), 0);
     }
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
+    export function SyncInvocation(callback: Callbacks.Generic, ...args: any) {
+        if (callback) callback(...args);
+    }
```
rugoncalves, line 10: Isn't this a bit overkill? Or you like the idea of isolating the method, in order to test if the handler exists?

glundgren93: I isolated the method, so I wouldn't have to change every call to the AsyncInvocation. But if you prefer, I can do that.

code/src/WijmoProvider/Features/ColumnSort.ts
glundgren93: The problem with this is that it is not guaranteed that the Grid will have an OnSort event. In these cases, the event will never get triggered.

rugoncalves (approved):

PR #201, opened 2022-02-15
ROU-3103 -  Created new ExtDate class and changed ToOSFormat

BenOsodrac:
This PR is for fixing the format method, which was not working correctly for date columns, as original conversion solution did not retrieve the correct binding for the column.

### What was done
* Created new ExtendedDate class, that extends the Date and adds an isDate property.
* Changed ToOSFormat method, so that it can accurately cycle the structure from the platform and format the date fields

### Test Steps
1. Open the screen
2. Click 'Json deserialize' button
3. The FeedbackMessage should return a date correctly format (no null date)

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/OSFramework/OSStructure/ExtendedDate.ts
+++ b/code/src/OSFramework/OSStructure/ExtendedDate.ts
@@ -1,0 +1,7 @@
+// Auxiliary class that helps us define which type of date comes from the platform data (date or datetime)
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+namespace OSFramework.OSStructure {
+    export class ExtendedDate extends Date {
+        public isDate: boolean;
+    }
+}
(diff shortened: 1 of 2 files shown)


code/src/OSFramework/OSStructure/ExtendedDate.ts
```diff
@@ -0,0 +1,7 @@
+// Auxiliary class that helps us define which type of date comes from the platform data (date or datetime)
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+namespace OSFramework.OSStructure {
+    export class ExtendedDate extends Date {
+        public isDate: boolean;
```
gnbm, line 5: I would add a comment here explaining the meaning/purpose of the isDate since it can cause readability issues in a class that extends Date @BenOsodrac @glundgren93

BenOsodrac: @gnbm We have a comment on the top. Are you saying that is not enough? Or you didn't saw it? :)

gnbm: I think it's enough for now

gnbm (approved):

PR #202, opened 2022-02-16
ROU-3103

BenOsodrac:
This PR is for fixing the format method, which was not working correctly for date columns, as original conversion solution did not retrieve the correct binding for the column.

### What was done
* Changed ToOSFormat method, so that it can accurately cycle the structure from the platform and format the date fields

### Test Steps
Open the screen
Click 'Json deserialize' button
The FeedbackMessage should return a date correctly format (no null date)

### Checklist
* [x] tested locally
* [x] documented the code


rugoncalves (approved):
Wow, that looks ingenious!

gnbm (approved):

PR #204, opened 2022-02-17
ROU-3087: Fix null date issue

glundgren93:
This PR fixes an issue where ArrangeData was returning wrong dates whenever we had Null Dates on a server that had a daylight saving.

### What was done
*  Only convert date to UTC if the date is not null


gnbm (approved):
LGTM :)

PR #208, opened 2022-02-23
ROU-3106: Arrange Data not working properly when using complex struct…

gnbm:
Fix an issue when handling more complex structures while obtaining the metadata for the generated JSON from the ArrangeData method.

### What was happening
* During a testing session, we found out that the Data Grid was throwing errors when the resulting JSON comes from a more complex structure like a structure with entities as an attribute
* We found out that the issues are inside the ArranjeData and that in these scenarios we’re not removing the prefix “ss“

### What was done
* Changed the method _cleanRecName_ to handle more complex structures when obtaining the metadata

### Test Steps
1. Create a screen with a grid with a data source from an advanced query that is appended to a list with a structure where we have an entity and another structure inside
2. Check that the information is loaded successfully
3. Click in **Deserialize**
4. Check that the data is correct and that the product name "My Personal Desktop" is displayed on a feedback message

### Screenshots
ROU3106_Preview

### Checklist
* [X] tested locally
* [X] documented the code


code/extension/DataGridUtils/Source/NET/obj/Release/DataGridUtils.csproj.FileListAbsolute.txt
```diff
@@ -0,0 +1,7 @@
+C:\Users\gna\Desktop\DataGridUtils\Source\NET\bin\OutSystems.NssDataGridUtils.dll
```
glundgren93, line 1: I don't think we should be committing files from obj here. What do you think?

glundgren93 (approved):

code/extension/DataGridUtils/Source/NET/DataGridUtils.cs
```diff
@@ -1,7 +1,7 @@
 
-namespace OutSystems.NssDataGridDataConvert {
+namespace OutSystems.NssDataGridUtils {
 
-	public class CssDataGridDataConvert: IssDataGridDataConvert {
+	public class CssDataGridUtils: IssDataGridUtils {
```
rugoncalves, line 4: Nice detail!

code/extension/DataGridUtils/Templates/NET/Interface.cs
```diff
@@ -22,6 +22,6 @@ public interface IssDataGridDataConvert {
 		/// <param name="ssDataJSON">JSON with the data and metadata for the data grid</param>
 		void MssConvertData2JSON(object ssData, out string ssDataJSON);
```
rugoncalves, line 23: Let's remove this method if it's deprecated? The extension is only used internally.

rugoncalves (approved):
Looks good - would just propose to remove the deprecated method.

code/extension/DataGridUtils/Templates/NET/Interface.cs
gnbm: will do that on the next release

PR #211, opened 2022-02-25
ROU-3105: Creating CustomizeExportingMessage API method

glundgren93:
This PR adds a default exporting message to the Grid and a new API method called CustomizeExportingMessage. 

### Test Steps
1. Open the context menu
2. Export to Excel
Expected: An exporting message saying "Your data is being exported" should appear during the file download and it should disappear after the file is downloaded.

-
1. Change ExportingMessage input to: "Custom exporting message".
2. Press CustomizeExportingMessage button. A success message should appear.
3. Export to Excel
Expected: An exporting message saying "Custom exporting message" should appear during the file download and it should disappear after the file is downloaded.

-
1. Toggle off the ShowMessage switch.
2. Export to Excel
Expected: No exporting message should appear.
-
1. Press CustomizeExportingMessage Invalid button. 
Expected: An error message should appear.


joselrio (approved):
💪

code/src/WijmoProvider/Features/Export.ts
```diff
@@ -45,10 +57,38 @@ namespace WijmoProvider.Feature {
...
+                '#' +
+                    gridId +
+                    ' .' +
+                    OSFramework.Helper.Constants.gridAutogeneratedPlaceholderCss
+            );
+            const loadingPlaceholderContent = `<div class="datagrid-loading heading4 OSInline" style="width: 100%; height: 100%;"><i class="icon fa-spin fa fa-spinner fa-1x"></i><div class="OSInline" style="margin-left: 10px;">${this._loadingMessage}</div></div>`;
+            const createdDivElem = document.createElement('div');
+            createdDivElem.className =
+                OSFramework.Helper.Constants.overlayExportFeedbackCss;
+            createdDivElem.innerHTML = loadingPlaceholderContent;
```
joanabpereira, line 72: Using innerHTML here will not pass if a security audit is performed on this code. The usage of innerHTML is dangerous and can lead to cross-site scripting. At the moment, there is a experimental technology that is used to parse and sanitize a string of HTML and then insert it into the DOM as a subtree of the element - Element.setHTML(). But as I said, it's experimental.
There is another way of going around this: https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString. Text is sanitized and transformed into HTML element (if you specify that way), all you have to do is to append that element to the one you need (createdDivElement.append(element))

code/src/WijmoProvider/Features/Export.ts
```diff
@@ -45,10 +57,38 @@ namespace WijmoProvider.Feature {
...
             return;
         }
 
+        public customizeExportingMessage(
+            exportingMessage: string,
+            showMessage = true
+        ): void {
+            if (exportingMessage !== '' && exportingMessage !== undefined) {
+                this._loadingMessage =
+                    OSFramework.Helper.Sanitize(exportingMessage);
```
joanabpereira: If we were to use the suggestion above, this would no longer be needed. But I would leave as is, just to be sure :)

joanabpereira (approved):

code/src/WijmoProvider/Features/Export.ts
gnbm: Nice catch @joanabpereira 
@glundgren93 Can you please review this?

gnbm: I'll create a task to improve this in the future and we'll move forward with this

PR #215, opened 2022-03-07
Getting the cell data for the correct type of column

gnbm:
This PR is to fix an issue when using parent binding and dropdowns,  where if we opened the dropdown without doing any changes, the associated column will also change and it shouldn’t since nothing changed.

### What was happening
* when using parent binding and dropdowns,  where if we opened the dropdown without doing any changes, the associated column will also change and it shouldn’t since nothing changed:
ParentBindingNoChangesIssue

### Test Steps
Validate parent binding when no changes on the dropdown options:
1. Open Page
2. In the 'Employee department' column, open the first dropdown and click on the same option.

Expected: 
- The value on the 'Employee' remains the same "

Validate parent binding with changes on the dropdown options:
1. Open Page
2. In the 'Employee department' column, open the first dropdown and click on a different option.

Expected: 
- The value on the 'Employee' is cleared"

### Checklist
* [X] tested locally
* [X] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/ValidationMark.ts
+++ b/code/src/WijmoProvider/Features/ValidationMark.ts
@@ -39,7 +39,12 @@ namespace WijmoProvider.Feature {
             e: wijmo.grid.CellRangeEventArgs
         ): void {
             const binding = s.getColumn(e.col).binding;
-            const newValue = s.getCellData(e.row, e.col, false);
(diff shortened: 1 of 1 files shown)


BenOsodrac (approved):

rugoncalves (approved):
This is a wonderful solution! Wow!

joselrio (approved):

PR #217, opened 2022-03-25
ROU-3217: OnCellValueChangeEvent of Data Grid not fired when two columns have the same binding

JoaoFerreira-FrontEnd:
This PR is for OnCellValueChangeEvent of Data Grid to be fired even when two columns have the same binding

### What was happening
* The OnCellValueChangeEvent was not being fired when I have two columns in the data grid binding to the same attribute in the database. I set a client action as the handler to the event on one of the columns. When I changed the value of the cell of that column, the client action does not get executed. However, when I remove the other column with the same binding, the event is fired and handled by the client action as it should be.

### What was done
* Change _triggerEventsFormColumn to use column index instead of column binding as input parameter

### Test Steps
1. Add a DataGrid to the screen
2. Config multiple columns with the same binding; 
3. Add custom events to some of the columns on OnCellValueChange;
4. Check if the event feedback is the expected

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/DirtyMark.ts
+++ b/code/src/WijmoProvider/Features/DirtyMark.ts
@@ -3,8 +3,7 @@ namespace WijmoProvider.Feature {
     export class DirtyMark
         implements
             OSFramework.Feature.IDirtyMark,
-            OSFramework.Interface.IBuilder
-    {
+            OSFramework.Interface.IBuilder {
         private _grid: Grid.IGridWijmo;
         private readonly _internalLabel = '__dirtyMarkFeature';
         private _metadata: OSFramework.Interface.IRowMetadata;
--- a/code/src/WijmoProvider/Features/DirtyMark.ts
+++ b/code/src/WijmoProvider/Features/DirtyMark.ts
@@ -3,7 +3,8 @@ namespace WijmoProvider.Feature {
     export class DirtyMark
         implements
             OSFramework.Feature.IDirtyMark,
-            OSFramework.Interface.IBuilder {
+            OSFramework.Interface.IBuilder
+    {
         private _grid: Grid.IGridWijmo;
         private readonly _internalLabel = '__dirtyMarkFeature';
         private _metadata: OSFramework.Interface.IRowMetadata;
--- a/code/src/WijmoProvider/Features/ValidationMark.ts
+++ b/code/src/WijmoProvider/Features/ValidationMark.ts
@@ -654,7 +654,7 @@ namespace WijmoProvider.Feature {
             // Triggers the events of OnCellValueChange associated to a specific column in OS
             this._triggerEventsFromColumn(
                 rowNumber,
-                column.provider.uniqueId,
+                column.uniqueId,
                 currValue,
                 currValue
             );
@@ -683,7 +683,7 @@ namespace WijmoProvider.Feature {
                         // Triggers the events of OnCellValueChange associated to a specific column in OS
                         this._triggerEventsFromColumn(
                             rowNumber,
-                            column.provider.uniqueId,
+                            column.uniqueId,
                             currValue,
                             currValue
                         );
--- a/code/src/WijmoProvider/Features/ValidationMark.ts
+++ b/code/src/WijmoProvider/Features/ValidationMark.ts
@@ -3,7 +3,8 @@ namespace WijmoProvider.Feature {
     export class ValidationMark
         implements
             OSFramework.Feature.IValidationMark,
-            OSFramework.Interface.IBuilder {
+            OSFramework.Interface.IBuilder
+    {
         private _grid: Grid.IGridWijmo;
         /** Internal label for the validation marks */
         private readonly _internalLabel = '__validationMarkFeature';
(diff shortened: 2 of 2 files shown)


code/src/WijmoProvider/Features/DirtyMark.ts
```diff
@@ -3,8 +3,7 @@ namespace WijmoProvider.Feature {
     export class DirtyMark
         implements
             OSFramework.Feature.IDirtyMark,
-            OSFramework.Interface.IBuilder
-    {
```
gnbm, line 7: @JoaoFerreira-FrontEnd The automation in Github are failing. Please check the feedback

gnbm (changes requested):

joselrio (approved):

code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -38,19 +38,27 @@ namespace WijmoProvider.Feature {
...
             );
             // The old value can be captured on the dirtyMark feature as it is the one responsible for saving the original values
             const oldValue = this._grid.features.dirtyMark.getOldValue(
                 e.row,
-                binding
+                column.binding
+            );
+            this._triggerEventsFromColumn(
+                e.row,
+                column.index,
```
glundgren93, line 58: We should be passing column uniqueId instead of column index.

glundgren93 (changes requested):

code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -304,7 +315,7 @@ namespace WijmoProvider.Feature {
                 );
                 this._triggerEventsFromColumn(
                     action.row,
-                    binding,
+                    action.col,
```
glundgren93, line 318: We should be passing column uniqueId instead of column index.

glundgren93 (changes requested):
Since we changed the behavior of the events, we should guarantee that it didnt cause any side effects as well.

code/src/WijmoProvider/Features/ValidationMark.ts
gnbm: @JoaoFerreira-FrontEnd This is still missing, right?

code/src/WijmoProvider/Features/ValidationMark.ts
gnbm: @JoaoFerreira-FrontEnd This is still missing, right?

gnbm (changes requested):

code/src/WijmoProvider/Features/ValidationMark.ts
JoaoFerreira-FrontEnd: No, this is done

gnbm (dismissed):

code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -229,19 +241,22 @@ namespace WijmoProvider.Feature {
...
             rowNumber: number,
-            binding: string,
+            columnUniqueID: string,
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             oldValue: any,
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             newValue: any
         ) {
-            const column = this._grid.getColumn(binding);
+            const column = this._grid
```
glundgren93, line 256: We should use: 
const column = this._grid.getColumn(columnUniqueID);

glundgren93 (changes requested):

gnbm (dismissed):

code/src/WijmoProvider/Features/ValidationMark.ts
JoaoFerreira-FrontEnd: done

code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -638,7 +654,7 @@ namespace WijmoProvider.Feature {
             // Triggers the events of OnCellValueChange associated to a specific column in OS
             this._triggerEventsFromColumn(
                 rowNumber,
-                column.provider.binding,
+                column.provider.uniqueId,
```
glundgren93, line 657: This should be column.uniqueId. The uniqueId is our property, not the providers.

code/src/WijmoProvider/Features/ValidationMark.ts
```diff
@@ -667,7 +683,7 @@ namespace WijmoProvider.Feature {
                         // Triggers the events of OnCellValueChange associated to a specific column in OS
                         this._triggerEventsFromColumn(
                             rowNumber,
-                            column.provider.binding,
+                            column.provider.uniqueId,
```
glundgren93, line 686: This should be column.uniqueId. The uniqueId is our property, not the providers.

glundgren93 (changes requested):

glundgren93 (dismissed):

glundgren93 (approved):

joselrio (approved):

PR #218, opened 2022-03-28
ROU-2898: Fixing multisort error on group panel

glundgren93:
This PR fixes a bug where multi sort was breaking when used on the group panel. This seemed to happen because whenever we removed the multi sort state on the group panel column, we were removing the sort description from wijmo’s sort array and wijmo’s code breaks because it expects two sort descriptions. So whenever it tries to access the second one, it throws the Cannot read properties of undefined (reading 'ascending'). 

### What was done
* Moved the deletion to the Sorted event and it seemed to work. Will run a regression to see if there is no side effect. 

### Test Steps
1. Drag Product Name into the group panel.
2. Pressing shift sort Product Category
3. Pressing shift sort Product Name
4. Not pressing shift, sort Product name again.

Expected: Product Name should be sorted desc and the sort from Product Category should be cleared.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,6 +27,8 @@ stages:
       - job: Build
         displayName: 'Build Job'
         timeoutInMinutes: 8
+        pool:
+          vmImage: 'windows-latest'
         steps:
           #Install dependencies
           - task: Npm@1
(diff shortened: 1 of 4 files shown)


gnbm:
@glundgren93 The automation from Azure pipelines are failing

gnbm (dismissed):

glundgren93:
> @glundgren93 The automation from Azure pipelines are failing

For some reason, there was an issue with the pipeline. Updated it according to the changes made on our tests and it seems to be working now. This PR will contain the pipeline updates as well.

BenOsodrac (approved):

PR #219, opened 2022-03-28
ROU-3183: Fix Issue when copy & paste a row that have child cells

joselrio:
This PR is for remove OnCellValueChange async trigger and set it definitely sync.

Since we had an IF checking if the action that trigger the OnCellValueChange was performed by a Filter it was synchrony triggered and changing it to be definitely synchrony triggered it will solve the Issue when copy & paste a row that have child cells (Parent Binding)


JoaoFerreira-FrontEnd (approved):

gnbm (approved):

glundgren93:
Please don't merge this yet, as we need #218 first, since we fixed an issue with azure.

PR #221, opened 2022-03-31
ROU-3145: Update SetCellData logic

gnbm:
This PR is for an improvement on the method SetCellData to have an extra param indicating if the on cell value change will be triggered or not.

### What was done
* Added an optional parameter (default true) to control via API the ability to trigger OnCellValueChange or not when using the SetCellData method

### Test Steps
1. Go to the test screen
2. Open the browser's console
3. Change the value of the first cell of the ProductName column
4. Checked that a loop with an error is displayed
5. Change the value of the first cell of the ProductCategory column
6. Checked that the value of the first cell of the ProductName column is now “New Name“ and that we only have one log on the console saying “Ran SetCellData Custom inside OnCellValueChange!“

### Screenshots
ROU3145_Test

Note: that the customer can achieve this by creating a client like this (equal to SetCellData but the last parameter = False): 
Capture

### Checklist
* [X] tested locally
* [X] documented the code
* [X] requires changes in OutSystems 
* [X] requires new sample page in OutSystems

### DIFF
--- a/code/src/GridAPI/Cells.ts
+++ b/code/src/GridAPI/Cells.ts
@@ -68,11 +68,13 @@ namespace GridAPI.Cells {
      * @param {string} gridID ID of the Grid.
      * @param {number} rowIndex Index of the row that contains the cells to be validated.
      * @param {string} columnID ID of the Column block in which the action of validation should be triggered.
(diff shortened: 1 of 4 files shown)


joselrio (approved):

glundgren93 (approved):
G

PR #230, opened 2022-04-25
ROU-3366: Adding another way to check if a column should be converted to Date/Datetime

glundgren93:
This PR adds another way to check if a column should be converted to Date/Datetime.

[Sample page](url)

### What was happening
* If we had an empty grid and added a a new row with a date format: 2020-20-12. As soon as we saved this using GetChangedLines, this column would be treated as if it were Date/Datetime, because when we were parsing the grid's data, we only looked for a regex match. So any column matching the date/datetime regex would be considered as such.
* Whenever we had an empty grid, added a new row and edited a Date/DateTime column, this value would not come on GetChangedLines. This was happening, because we were not considering empty cells on the ToJSONFormat method.

### What was done
* Created Column <Key, Type> map in order to use on setData.
* Created map from metadata as well.
* These maps will add an extra protection when parsing the JSON, as we should only consider date/dateTime on these respective columns.
* Adjusted ToJSONFormat to consider empty values as well.

### Test Steps

----------------------------------------------------
Column tests 

1. Add row on ArrangeDataColumn Grid
2. Insert 2020-20-12 on the email column
3. Insert any date on DateOfBirth column.
4. Press save all on ArrangeDataColumn Grid

Expected: Both grids on this page should be updated and contain inserted values and no dirty marks.

5. Edit DateOfBirthColumn on ArrangeDataColumn Grid.
6. Press save all on ArrangeDataColumn Grid.
7. Edit DateOfBirthColumn on JSONSerialize Column Grid.
6. Press save all on ArrangeDataColumn Grid.

Expected: Both grids on this page should be updated and contain inserted values and no dirty marks.

9. Press RemoveRow
10. Press save all on ArrangeDataColumn Grid

Expected: Both grids on this page should be empty

----------------------------------------------------
Autogen tests

1. Add row on ArrangeDataAutogen Grid
2. Insert 2020-20-12 on the Name column
3. Insert any date on Created column.
4. Press save all on ArrangeData Autogen 

Expected: Grid should be updated and contain inserted values and no dirty marks.

5. Edit DateOfBirthColumn on ArrangeDataAutogen   Grid.
6. Press save all on ArrangeDataAutogenGrid.

Expected:  Grid should be updated and contain inserted values and no dirty marks.

9. Press RemoveRow
10. Press save all on ArrangeDataAutogenGrid

Expected:  Grid should be empty

### DIFF
--- a/code/src/OSFramework/Grid/AbstractDataSource.ts
+++ b/code/src/OSFramework/Grid/AbstractDataSource.ts
@@ -53,7 +53,7 @@ namespace OSFramework.Grid {
                 };
 
                 const handleValue = (value) => {
-                    if (match(value, regex.datetime)) {
+                    if (match(value, regex.datetime) && type === 'DateTime') {
                         return new Date(
                             Date.UTC(
                                 +m[1],
@@ -64,7 +64,7 @@ namespace OSFramework.Grid {
                                 +m[6]
                             )
                         );
-                    } else if (match(value, regex.date)) {
+                    } else if (match(value, regex.date) && type === 'Date') {
                         return new Date(+m[1], +m[2] - 1, +m[3]);
                     } else if (value === '') {
                         return undefined;
--- a/code/src/OSFramework/Grid/AbstractDataSource.ts
+++ b/code/src/OSFramework/Grid/AbstractDataSource.ts
@@ -64,7 +64,10 @@ namespace OSFramework.Grid {
                                 +m[6]
                             )
                         );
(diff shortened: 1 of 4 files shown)


JoaoFerreira-FrontEnd (approved):

BenOsodrac (approved):

gnbm (approved):
Amazing work on this 🚀

joselrio (dismissed):

PR #235, opened 2022-05-11
ROU-3428: Fix column picker on autogen

glundgren93:
This PR fixes column picker on autogenerated grids

### What was happening
* Bug introduced on #227 as we had changed the column picker creation to use our columns instead of Wijmo's.

### What was done
* Added a condition to check if grid is autogenerated or not and use proper columns.

### Test Steps
1. Open column picker

Expected: All columns should be there.


gnbm (approved):

rugoncalves (approved):
LGTM!

PR #245, opened 2022-05-20
ROU-2905: [DataGrid] - Column header/cells word wrap 

JoaoFerreira-FrontEnd:
This PR is for an API that allows the developer to set the column with wordwrap.

### What was done
* Create API SetColumnWordWrap;

### Test Steps
1. Add the API to a flow;
2. Check if the column text changed;

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/code/src/WijmoProvider/Features/Styling.ts
+++ b/code/src/WijmoProvider/Features/Styling.ts
@@ -117,8 +117,8 @@ namespace WijmoProvider.Feature {
             const column = this._grid.getColumn(columnID);
             if (column) {
                 this._grid.provider.columns.find(
-                    (column) =>
-                        column.index ===
+                    (element) =>
+                        element.index ===
                         this._grid.getColumn(columnID).provider.index
                 ).wordWrap = value;
             } else {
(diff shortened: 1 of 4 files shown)


code/src/WijmoProvider/Features/Styling.ts
```diff
@@ -111,5 +111,21 @@ namespace WijmoProvider.Feature {
                 );
             }
         }
+
+        public setColumnWordWrap(columnID: string, value: boolean): void {
+            // validate if column exists
+            const column = this._grid.getColumn(columnID);
+            if (column) {
```
glundgren93, line 118: I believe you can just do column.provider.worWrap = value;

glundgren93 (changes requested):

code/src/WijmoProvider/Features/Styling.ts
JoaoFerreira-FrontEnd: I'll do that

glundgren93 (approved):

PR #250, opened 2022-05-25
ROU-3475: Fixing issue with GetChangedLines

glundgren93:
This fixes an issue where GetChangedLines was returning edited lines even though the row had its original value.

### What was happening
* Since we add rowMetadata on the grid's dataItem, wijmo can't tell if the dataItem has been changed or not, because for them the dataItem is always different from the original value.

### What was done
* Created a method that tells if Grid has dirty marks and uses that on GetChangedLines.

### Test Steps
1. Change the Product Category on the first cell.
2. Change back to the original value.
3. Press GetChanged Lines.

Expected: Should return empty values.

### DIFF
--- a/code/src/OSFramework/Feature/IDirtyMark.ts
+++ b/code/src/OSFramework/Feature/IDirtyMark.ts
@@ -1,6 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 namespace OSFramework.Feature {
     export interface IDirtyMark {
+        isGridDirty: boolean;
         clear(): void;
         clearByRowKeys(rowKeys: Array<string>): void;
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


code/src/OSFramework/Feature/IDirtyMark.ts
```diff
@@ -8,6 +8,7 @@ namespace OSFramework.Feature {
         clearPropertyInRowByKey(key: string): void;
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         getOldValue(rowNumber: number, binding: string): any;
+        isGridDirty(): boolean;
```
BenOsodrac, line 11: @glundgren93 congrats, best method name ever, I couldn't do better myself!

code/src/WijmoProvider/Features/DirtyMark.ts
```diff
@@ -244,6 +244,12 @@ namespace WijmoProvider.Feature {
             );
         }
 
+        public isGridDirty(): boolean {
+            return this._grid.provider.itemsSource.sourceCollection.some(
+                (row, index) => this._isDirtyRow(index)
```
BenOsodrac, line 249: I think the row parameter can be removed. Its throwing a code-smell also

JoaoFerreira-FrontEnd (approved):

PR #276, opened 2022-08-08
ROU-3667: Fixing changeProperty

glundgren93:
This PR fixes a bug where changing column properties would break columns that have a conditional format.

### Test Steps
1. Change header name

Expected: Header name of the first column should change and no error should be thrown.

### DIFF
--- a/code/src/OSFramework/OSStructure/ColumnProperties.ts
+++ b/code/src/OSFramework/OSStructure/ColumnProperties.ts
@@ -1,13 +1,13 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 namespace OSFramework.OSStructure {
     export enum ColumnProperties {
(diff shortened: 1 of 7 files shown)


code/src/OSFramework/OSStructure/ColumnProperties.ts
```diff
@@ -0,0 +1,13 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+namespace OSFramework.OSStructure {
+    export enum ColumnProperties {
```
gnbm, line 3: @glundgren93 Just to start being consistent with other assets like OS UI, please order the options alphabetically

gnbm (changes requested):

gnbm (approved):

joselrio (approved):

PR #277, opened 2022-08-09
ROU-3674: Fixing AddRows

OS-giulianasilva:
This PR is for add a new validation for when the number of added row does not fit in the current grid page.

### What was happening
* The added rows were exceeding to the next grid page when it exceeds the RowsPerPage number.

### What was done
* Added a new validation in _validateAddNewRow method
* Verifies if the amount of new rows plus the selected row index is greater than the max rows per page. If true, throw an error, otherwise, do nothing.
* New error message for this case.

### Test Steps
1. Click on the last row of the page and try to add 2 new rows. Expect to throw an error.
2. Click on the first row and try to add a greater amount of rows than the page limit. Expect to throw an error.
3. Click on the first row and try to add the same amount of rows as the page limit. Expect to add all new rows.

### Checklist
* [x] tested locally
* [ ] documented the code


glundgren93 (approved):
Nice!

gnbm (approved):
@OS-giulianasilva Great work 👍

glundgren93:
There are some tests failing. Let's have a look at them before merging.

glundgren93 (approved):

gnbm (approved):

PR #285, opened 2022-09-13
ROU-3673: Improving undo add rows performance

glundgren93:
This PR improves undo add rows performance.

### What was happening
* We were using collectionView.remove, which iterates over array to find item index and them eventually uses splice to remove it.

### What was done
* Changed the itemsAdded removal to splice directly, since we can find out the starting index from beginning.


gnbm (approved):
Amazing work @glundgren93

BenOsodrac (approved):

PR #298, opened 2022-10-21
ROU-3885: Fix getTime bug when the date is in string format

OS-giulianasilva:
This PR is for fix the bug when changing the value of a Date Column that initially has a DateTime type data assigned to it. Same issue was also happening when calling GetChangedRosw action.

### What was happening
* When passing DateTime attributes in a Date column and vice versa, we were getting the error `date.getTime() is not a function`.
* The GetChangedRows action was not returning the changed rows when the Datetime column with a date attribute gad it values changed.

### What was done
* Inserted an assertion that check if the value that comes from a Date and DateTime columns is a string. If it is, so we convert it to a Date format.

### Test Steps
1. Go to test page.
2. Open the Dev Tools console tab
3. Change the date of the first column
4. Check that no error has appeared in the console
5. Click in GetChangedRows button
6. Check if the new value for that columns was the expected
7. Repeat it for the other columns
Note: GetChangedRows returns the server Date or DateTime.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
+++ b/src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
@@ -119,12 +119,7 @@ namespace OSFramework.DataGrid.Grid {
                     (columnsType.get(key) === 'Date' || dateColumns.has(key)) &&
                     object[key]
                 ) {
-                    const dt = object[key] as Date;
-                    object[key] = new Date(
-                        dt.getTime() - dt.getTimezoneOffset() * 60000
-                    )
-                        .toISOString()
-                        .substring(0, 10);
+                    object[key] = OSFramework.DataGrid.Helper.ToOSDate(object[key])
                 }
             });
         };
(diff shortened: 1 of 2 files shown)


src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
```diff
@@ -119,12 +119,7 @@ namespace OSFramework.DataGrid.Grid {
                     (columnsType.get(key) === 'Date' || dateColumns.has(key)) &&
                     object[key]
                 ) {
-                    const dt = object[key] as Date;
-                    object[key] = new Date(
-                        dt.getTime() - dt.getTimezoneOffset() * 60000
-                    )
-                        .toISOString()
-                        .substring(0, 10);
+                    object[key] = DataGrid.Helper.ToOSDate(object[key]);
```
JoaoFerreira-FrontEnd, line 122: It will result with dates that aren't on the same timezone has the server?

BenOsodrac (approved):

JoaoFerreira-FrontEnd (approved):

PR #300, opened 2022-11-11
ROU-3936: Fix GetChangedLines not working when not on the first page

OS-giulianasilva:
This PR is for fix the issue with the GetChangedLines returning an empty array when not in the first page, even with edited lines in the page, and when filter and sort were applied.

### What was happening
* In src\Providers\DataGrid\Wijmo\Grid\ProviderDataSource.ts:
   * In _getDirtyEditedItems():
       * itemsSource.sourceCollection.findIndex() returns the index based on all data, not only in the current view.
       * isRowDirty() method was returning false when the rowIndex >= number of rows in the view, because the _hasMetadata inside the _isDirtyRow() method only validates the metadata of the rows in the current view.

### What was done
* Because of other issues found during the development of this task, it was necessary to rollback a previous fix related to GetChangedLines (ROU-3592). As a consequence, the bug that the fix solved, is happening again but in a different way.
  * In ProviderDataSource.ts:
     * In getChanges method:
       * The editedLinesJSON is now using the itemsource.itemsEdited from Wijmo again.
       * The isGridDirty validation was removed, so the GetChangedLines return the editedLines even when the view is not dirty.
  * In ValidationMark.ts:
    * In build method:
      * Added new handler to the cellEditEnding event.
      * The handler prevents changes from null to "" to be considered an effective change.
      * It partially fix the ROU-3592 (it still occurs in a very specific case related to calculated columns).
 
### Test Steps
- Test 1:
  1. Go to APIData test page.
  2. Edit any row in the second page of the grid.
  3. Go back to the first page.
  4. Click on GetChangedLines button.
  Expected: The displayed message shows the edited lines.
  5. Edit any row in the first page of the grid.
  6. Click on GetChangedLines button.
  Expected: The displayed message shows all the edited lines.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
@@ -35,16 +35,17 @@ namespace Providers.DataGrid.Wijmo.Feature {
         /**
          * Handler for the CellEditEnding.
          */
-        private _cellEditBeforeEndingHandler( s: wijmo.grid.FlexGrid,
+        private _cellEditBeforeEndingHandler(
+            s: wijmo.grid.FlexGrid,
             e: wijmo.grid.CellRangeEventArgs
         ): void {
             // get old and new values
-             let oldVal = s.getCellData(e.row, e.col, false) ?? '',
-                 newVal = s.activeEditor.value ?? '';
-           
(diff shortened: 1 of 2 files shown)


joselrio (approved):
💪

bmarcelino-fe (approved):

PR #310, opened 2022-12-20
ROU-4011: Fix parent dropdown not erasing the child when returning to its original value

OS-giulianasilva:
This PR is for fix parent dropdown not erasing the child when returning to its original value

### What was happening
* The parent dropdown handler was not erasing the child value when oldValue === newValue

### What was done
* Removed the code that erases the dropdown child cell from the if that verifies if the oldValue is different than the newValue.
* Now it also erases the values when the oldValue is equal to the newValue.

### Test Steps
1. Go to the DropdownDependency test page 
2. Change the parent dropdown of the first row to different value
3. Change the child dropdown of the first row to a different value
4. Change the parent dropdown of the first row to the original value
5. Check if the child dropdown of the first row is empty, has a dirty mark and is invalid.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Columns/DropdownColumn.ts
+++ b/src/Providers/DataGrid/Wijmo/Columns/DropdownColumn.ts
@@ -140,83 +140,87 @@ namespace Providers.DataGrid.Wijmo.Column {
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             newValue: any
         ): void {
-            // only set to blank if there is a different value
             if (
                 oldValue !== newValue &&
                 oldValue.toString() !== newValue.toString()
             ) {
-                const column = this.grid.getColumn(columnID);
-                const currentValue = this.grid.provider.getCellData(
-                    rowNumber,
-                    this.provider.index,
-                    false
-                );
-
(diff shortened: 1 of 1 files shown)


src/Providers/DataGrid/Wijmo/Columns/DropdownColumn.ts
```diff
@@ -140,83 +140,87 @@ namespace Providers.DataGrid.Wijmo.Column {
...
-                        // add new child action into existing parent action in order
-                        existingUndoAction.addChildAction(
-                            new GridEditAction(
-                                this.grid,
-                                new wijmo.grid.CellRangeEventArgs(
-                                    this.grid.provider.cells,
-                                    cellRange
-                                )
+                // only add child undo action if it doesn't already exist. we don't want duplicated actions
+                if (!existingEditActionForColRow) {
```
JoaoFerreira-FrontEnd, line 187: change it to (existingEditActionForColRow === false) please

JoaoFerreira-FrontEnd (changes requested):

JoaoFerreira-FrontEnd (approved):

bmarcelino-fe (approved):

PR #335, opened 2023-05-19
Changed ClearFilter to allow triggering OnFiltersChange event

gnbm:
This PR is to implement a way to trigger the OnFiltersChange event when using the ClearFilter.

### What was happening
* When using the Filters API, only the ClearFilter was not triggering the OnFiltersChange event.

### What was done
* Changed ClearFilter to allow triggering OnFiltersChange event

### Test Steps
1. Go to GridReactive_Automation_ROU4340/API_Filter
2. Clicked in Filter by Value
3. Checked that the Server Side Grid was filtered and the OnFiltersChange event triggered
4. Clicked in Filter Clear
5. Checked that the Server Side Grid got the filters cleared and the OnFiltersChange event triggered
6. Do the same using the Grid filters to filter and clear, checking that the OnFiltersChange event triggered

### Checklist
* [X] tested locally
* [X] documented the code
* [X] requires changes in OutSystems (if so, provide a module with changes)

### DIFF
--- a/src/OSFramework/DataGrid/Configuration/Column/EditorConfigDate.ts
+++ b/src/OSFramework/DataGrid/Configuration/Column/EditorConfigDate.ts
@@ -10,7 +10,10 @@ namespace OSFramework.DataGrid.Configuration.Column {
         public min: Date;
 
         // eslint-disable-next-line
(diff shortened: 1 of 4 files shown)


JoaoFerreira-FrontEnd (approved):

rugoncalves (approved):
Looking good!

PR #336, opened 2023-05-22
Update ToolTip.ts

gnbm:
This PR is for adding validation to check if we do have data available, for instance, while using filters that make the Grid without results.

### What was happening
* When applying a filter that showed no results, while hovering the Grid we got errors thrown in the Developer's Tools console.

### What was done
* Added validation to check if we do have data available, for instance, while using filters that make the Grid without results.

### Test Steps

**Test Case 1**:
1. Go to _OutSystemsDataGridSample_ROU4335/Overview_
2. Mouse Hover the Grid cells, including some with text truncated
3. Check that the tooltip is displayed and no errors are thrown in the Dev Tools
4. Mouse Hover the Grid Headers
5. Check that the tooltip is displayed and no errors are thrown in the Dev Tools

**Test Case 2 - Filtered List**:
1. Go to _OutSystemsDataGridSample_ROU4335/Overview_
2. Filter the user name column by condition, being condition Equal to “ZZZZ“
3. Check that the Grid is empty
4. Mouse Hover the Grid cells, including some with text truncated
5. Check that the tooltip is displayed and no errors are thrown in the Dev Tools
6. Mouse Hover the Grid Headers
7. Check that the tooltip is displayed and no errors are thrown in the Dev Tools

### Screenshots
* Issue:

IssueGridOnHoverHeader

* After the fix: 

IssueGridOnHoverHeaderFixed

### Checklist
* [X] tested locally
* [X] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ToolTip.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ToolTip.ts
@@ -28,11 +28,14 @@ namespace Providers.DataGrid.Wijmo.Feature {
                 cellType === wijmo.grid.CellType.Cell ||
                 cellType === wijmo.grid.CellType.ColumnFooter
             ) {
-                this._setCellToolTip(
-                    _currTarget,
-                    ht.getColumn().binding,
-                    ht.row
-                );
+                //Check if we do have data available, for instance while using filters that make the Grid without results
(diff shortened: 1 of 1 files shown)


BenOsodrac (approved):

rugoncalves (approved):
Looking good! Let's just have a full regression on this one! :)

PR #337, opened 2023-05-24
ROU-4319: adding the ability to block autogenerated grids

rugoncalves:
This PR is for adding the configuration that will allow to block the auto generation of the grid via configuration.
This feature is relevant for ODC, since auto generated grids will not be allowed to be used, since the a extension to obtain the metadata will not be available, and with that a degradation of the experience.

### What was happening
* When using the wijmo auto generate of the columns the experience is tremendously downgraded, both DEV and runtime.
* It is not possible to make editable, because the data will not match the one coming from OutSystems.

### What was done
* Added a new config that will tell if this feature is to be blocked or not: `allowAutoGenerate` in the structure `Internal_Configs`
* In ODC will have the value will be `false`, while in O11 `true`;
* The config is being tested in the private method `_autoGenCol` of the class `AbstractGrid`
* A new error message was added to the framework;

### Disclaimer
For testing purposes, at the moment a **new** input was added to the Grid block, that will be **promptly** removed when the merge is done to the trunk.

### Test Steps
1. Go to the test page provided
2. Change the variables and see the impact:
image
3. When using the first config (`Allow columns auto generation`) to false, then an error should appear.

### Screenshots
grid-odc

### Checklist
* [x] tested locally
* [x] documented the code


src/OSFramework/DataGrid/Configuration/IConfigurationGrid.ts
```diff
@@ -6,6 +6,12 @@ namespace OSFramework.DataGrid.Configuration {
      */
     export interface IConfigurationGrid
         extends IConfiguration<DataGrid.Types.IGridProviderConfigs> {
+        /**
+         * Indicates if the grid should be allowed to be autogenerated (at all) or not.
+         * In ODC, due to the inexistance of the
```
gnbm, line 11: @rugoncalves Please check these GitHub actions feedback

src/OSFramework/DataGrid/Configuration/IConfigurationGrid.ts
```diff
@@ -6,6 +6,12 @@ namespace OSFramework.DataGrid.Configuration {
      */
     export interface IConfigurationGrid
         extends IConfiguration<DataGrid.Types.IGridProviderConfigs> {
+        /**
+         * Indicates if the grid should be allowed to be autogenerated (at all) or not.
+         * In ODC, due to the inexistance of the 
+         */
+        allowAutoGeneratedGrid: boolean;
+        
         /**
```
gnbm, line 15: Same as bove

gnbm (changes requested):

joselrio (commented):
despite @gnbm mentioned already, LGFM.

BenOsodrac (approved):

BenOsodrac (approved):

PR #340, opened 2023-06-06
ROU-4367: Fixed issue in moveToPage error handling

gnbm:
This PR is for fixing an issue in the `moveToPage()` method that was always throwing an error.

### What was happening
* Even in the case of a successful call, the client action **SetCurrentPage** was throwing an error.

### What was done
* Fixed the condition that was throwing the error.

### Test Steps
1. Go to GridReactive_Automation_ROU4367/API
2. Add the page to go in the input set as 5
3. Click in "Set Current Page"
4. Check that the page changed and no error was returned

### Screenshots
image

### Checklist
* [X] tested locally
* [X] documented the code


OS-giulianasilva (approved):

rugoncalves (approved):
I typically prefer more visual test `=== false` but it's a personal preference!

PR #350, opened 2023-06-26
ROU-4388: Improve header's accessibility

OS-giulianasilva:
This PR is for set Row and column headers to be accessible via keyboard.

### What was happening
* The Row and Column headers were not accessible via keyboard by default.

### What was done
* Set the headersFocusability property to "All" by default allowing the Row and Column headers to be accessible.

### Test Steps
1. Go to a page with a Grid with accessibility feature enabled
2. Check that it is possible to access the Row and Column headers via keyboard.

### Checklist
* [x] tested locally
* [x] documented the code


BenOsodrac (approved):

gnbm (approved):
Well done @OS-giulianasilva

PR #355, opened 2023-07-03
Update typedoc to version `0.24.8`

rugoncalves:
This PR is for updating **Typedoc (dev) dependency version**, since version `0.24.7` was breaking the PR pipeline.

### What was happening
* The version `0.24.7` of **typedoc** was breaking the pipeline
* The problem was related with this specific version of typedoc (the root causes are unknown)

### What was done
* Updated the version of typedoc to `0.24.8`

### Checklist
* [x] tested locally
* [ ] documented the code


package.json
```diff
@@ -25,8 +25,8 @@
     "eslint-plugin-prettier": "^3.4.0",
     "prettier": "^2.3.0",
     "prettier-eslint": "^12.0.0",
-    "typedoc": "^0.24.7",
```
gnbm, line 28: @rugoncalves Probably this will also happen on other assets

gnbm (approved):

package.json
rugoncalves: possibly... I'll check.

OS-giulianasilva (approved):

PR #357, opened 2023-07-13
ROU-4304: Revert empty strings being converted in undefined

OS-giulianasilva:
This PR is for revert empty strings being converted in undefined.

### What was happening
* OS Date/DateTime fields are sent as strings, and we should parse it to the correct data type. When they are empty strings, we convert it to undefined.
* The problem is that we are doing it for all data that are empty strings, which is not necessary

### What was done
* Changed the ToJSONFormat function of AbstractDataSource file to only convert to only convert empty string in undefined in two cases:
    * When it is Date or DateTime columns data
    * When it is an unknown data type, because it can or cannot be Date/DateTime, so we convert it to avoid errors.

### Test Steps
1. Go to a page witha Grid to the page containing some data with empty values in Text, Date and DateTime columns.
2. Call the GetRowData Client Action for the desired RowNumber
3. Check that the Text Column data are returned as empty, but the Date and DateTime data are undefined.

### Screenshots
Before:
image
After:
image

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
+++ b/src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
@@ -18,7 +18,7 @@ namespace OSFramework.DataGrid.Grid {
      * Responsible for parsing string fields to its correct data type
      *
      * @param data A string of the JSON data object
-     * @example OS Date fields is send as strings, and we should parse it to Date, the is done for Datetime fields.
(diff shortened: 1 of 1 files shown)


joselrio (approved):

rugoncalves (approved):
LGTM

PR #358, opened 2023-07-18
ROU-4414: Review Context Menu lifecycle for auto-generated Grids

OS-giulianasilva:
This PR is for...

[Sample page](url)

### What was happening
* In autogenerated Grids, we were not covering cases when the column binding is different from the type "Entity.Field" when checking for new columns in `checkForNewColumns` function in `AbstractGrid`.
* Because of that, when the Grid has a binding in the format "Field" instead of "Entity.Field" for example, and the data source is refreshed, it accuses to have new columns added and the initialized events run again without need.

### What was done
* Fix `checkForNewColumns` function in `AbstractGrid` to cover these missing cases.
* Now, we check if the `metadata[source]` is an object or not. If it is the binding is in the format "Entity.Field", otherwise it is in the format "Field".
`
### Test Steps
1. Go to a page with an autogenerated Grid.
2. The data source should have bindings in the format "Field" and "Entity.Field"
3. Refresh the data source 
4. Check that the console does not have "Remove column ..." logs.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/OSFramework/DataGrid/Grid/AbstractGrid.ts
+++ b/src/OSFramework/DataGrid/Grid/AbstractGrid.ts
@@ -156,7 +156,7 @@ namespace OSFramework.DataGrid.Grid {
                     hasColumns = newColumns.every((column) => {
                         return columns.indexOf(`${source}.${column}`) !== -1;
                     });
-                } 
+                }
                 // If not an object, we assume it is a string, meaning that the column binding is the "source" itself
                 // So we check if this is a new column
                 else {
--- a/src/OSFramework/DataGrid/Grid/AbstractGrid.ts
+++ b/src/OSFramework/DataGrid/Grid/AbstractGrid.ts
@@ -149,10 +149,19 @@ namespace OSFramework.DataGrid.Grid {
                 (col) => col.config.binding
             );
             return Object.keys(metadata).some((source) => {
-                const newColumns = Object.keys(metadata[source]);
-                hasColumns = newColumns.every((column) => {
-                    return columns.indexOf(`${source}.${column}`) !== -1;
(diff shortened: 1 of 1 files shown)


src/OSFramework/DataGrid/Grid/AbstractGrid.ts
```diff
@@ -149,10 +149,19 @@ namespace OSFramework.DataGrid.Grid {
                 (col) => col.config.binding
             );
             return Object.keys(metadata).some((source) => {
-                const newColumns = Object.keys(metadata[source]);
-                hasColumns = newColumns.every((column) => {
-                    return columns.indexOf(`${source}.${column}`) !== -1;
-                });
+                const isObject = typeof metadata[source] === 'object';
```
rugoncalves, line 152: This constant, can be removed, and the test done directly in the id.

src/OSFramework/DataGrid/Grid/AbstractGrid.ts
```diff
@@ -149,10 +149,19 @@ namespace OSFramework.DataGrid.Grid {
...
+                // If it is an object, let's iterate into the keys to check if there are new columns
+                if (isObject) {
+                    const newColumns = Object.keys(metadata[source]);
+                    hasColumns = newColumns.every((column) => {
```
rugoncalves, line 156: Use `some` instead of `every` in the inner loop because we only need to find one new column to return true and terminate the iteration early, which makes it more efficient.

rugoncalves (changes requested):
Small improvement

bmarcelino-fe (approved):

PR #367, opened 2023-07-28
ROU-4472: Improve the SetColumnWordWrap to make the DynamicHeight  being applied to the header

gnbm:
This PR is to improve the _SetColumnWordWrap_ to make the _DynamicHeight_ applied to the header.

### What was happening
* When setting _DynamicHeight_ in the client action _SetColumnWordWrap_ it only applies to the cells with data, not to the headers:

ROU-4472_Before

### What was done
* Now, when setting `DynamicHeight = True` the client action _SetColumnWordWrap_ the header will also receive the same behaviour.

### Test Steps
1. Go to test page APIStyling
2. Reduce the Product Name column width
3. Click on “SetColumnWordWrap“
4. Checked that both data cells and header have the height adapted to their content 

### Screenshots

ROU-4472_AfterImprov

### Checklist
* [X] tested locally
* [X] documented the code

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,2483 @@
 {
   "name": "outsystems-datagrid",
-  "version": "2.12.1",
-  "lockfileVersion": 1,
+  "version": "2.13.0",
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "outsystems-datagrid",
+      "version": "2.13.0",
+      "license": "BSD-3-Clause",
+      "dependencies": {
+        "@types/lodash": "^4.14.194",
+        "typescript": "^5.0.4"
+      },
+      "devDependencies": {
+        "@typescript-eslint/eslint-plugin": "^4.23.0",
+        "@typescript-eslint/parser": "^4.23.0",
+        "eslint": "^7.26.0",
+        "eslint-config-prettier": "^7.0.0",
+        "eslint-plugin-prettier": "^3.4.0",
+        "prettier": "^2.3.0",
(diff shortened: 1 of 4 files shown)


OS-giulianasilva (approved):

rugoncalves (commented):
Shouldn't this only enter together with the other task?

docs/assets/highlight.css
rugoncalves, line 1: Was changing this file really needed? :)

docs/assets/search.js
rugoncalves, line 1: Was this change needed?

docs/assets/style.css
rugoncalves, line 1: These changes look dubious. Will they work correctly in css, I mean replacing double quotes by single quotes.

rugoncalves (approved):
Upon discussion, let's move forward.

docs/assets/highlight.css
gnbm: @rugoncalves This was the result of running lintfix that was getting blocked on the PR automation

docs/assets/search.js
gnbm: @rugoncalves This was the result of running lintfix that was getting blocked on the PR automation.

docs/assets/style.css
gnbm: @rugoncalves This was the result of running lintfix that was getting blocked on the PR automation. But I'm also not sure about the CSS part

gnbm: Maybe we should rollback this on?

gnbm: @rugoncalves Just tested it and it works

PR #368, opened 2023-08-03
Improving performance api

glundgren93:
This PR would be the first step on reducing some verbosity on DataGrid's public API methods. It creates an auxiliary function that would execute and measure the performance of a given function. Only changed GridManager, as I understand changing all APIs would make this PR tough to read/test.

### What was happening
* Currently our public APIs are very verbose. For each API method, we call Perfomance.SetMark and Performance.GetMeasure, while repeating the function's name a few times.

### What was done
* Created an auxiliary function called MeasurePerformance that receives a functionName and the function itself and returns the output of the function, while measuring its performance.

### Test Steps
1. Open this [sample](https://outsystemsui.outsystems.com/OutSystemsDataGridSample/SearchingData)
2. To enable performance measurements, we must call this method `OutSystems.GridAPI.Performance.SetDebug(true)`.
3. Then call OutSystems.GridAPI.Performance.GetAllMeasures()
4. Expected: You should see GridManager methods registered there.

### Screenshots
Before change:
image

After change:
image

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/OutSystems/GridAPI/Auxiliary.ts
+++ b/src/OutSystems/GridAPI/Auxiliary.ts
@@ -64,6 +64,12 @@ namespace OutSystems.GridAPI.Auxiliary {
         return OSFramework.DataGrid.Helper.GenerateHashCode(str);
     }
 
+    /**
+     * Receives a function and its name. Executes it and measures it properly.
+     * @param functionName Name of the function that will be measured
+     * @param fn Function that will be measured and executed
+     * @returns Output of the received function
+     */
     export function MeasurePerformance<T extends (...args: any[]) => any>(
         functionName: string,
         fn: T
--- a/src/OutSystems/GridAPI/Auxiliary.ts
+++ b/src/OutSystems/GridAPI/Auxiliary.ts
@@ -63,6 +63,25 @@ namespace OutSystems.GridAPI.Auxiliary {
     export function GetHashCode(str: string): number {
         return OSFramework.DataGrid.Helper.GenerateHashCode(str);
     }
+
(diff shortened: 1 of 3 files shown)


gnbm (approved):
@glundgren93 Thank you for the contribution. Feel free to continue them coming 💪🏻🚀

rugoncalves (approved):
Oh yeah! @glundgren93 is rolling!

src/OutSystems/GridAPI/Auxiliary.ts
```diff
@@ -63,6 +63,32 @@ namespace OutSystems.GridAPI.Auxiliary {
...
     }
+
+    /**
+     * Receives a function and its name. Executes it and measures it properly.
+     * @param functionName Name of the function that will be measured
+     * @param fn Function that will be measured and executed
+     * @returns Output of the received function
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    export function MeasurePerformance<T extends (...args: any[]) => any>(
```
rugoncalves, line 74: actually you could replace `any` with `unknown`, no @glundgren93?

rugoncalves (approved):

src/OutSystems/GridAPI/Auxiliary.ts
glundgren93: I tried that, but it raised another error, so I opted for this instead.

BenOsodrac (approved):

gnbm:
@glundgren93 Thank you for your contribution!!

PR #375, opened 2023-09-12
ROU-4519: Issue with dropdown column with dependency triggering OnCellValueChange event twice

OS-giulianasilva:
This PR is for fix the dropdown child triggering the OnCellValueChange more than once.

### What was happening
- When the Grid has a dropdown column with dependency, the OnCellValueChange event was triggered twice when the parent cell value changed.
- 

### What was done
- Add validation to check if currentValue is different than empty when adding the child undo action and triggering the OnCellValueChange event.
   - If the currentValue is already empty, we don't need to set the cell value to empty again and add it to the undo stack.
- Change the way we get the last parent cell's undo action on the undo stack.
   - We want to ensure that the child's undo action is added to the last parent's undo action in the undo stack. 
- Disabled the trigger in the validateCell method to trigger the OnCellValueChange event just once.
   - The onCellValueChange is already explicitly being triggered in the _parentCellValueChangeHandler, We don't need to do this twice.
- Trigger the OnCellValueChange with the original value instead of currentValue.
   - The OnCellValueChange event must return the original value and not the previous cell value.
- Added an else statement in ValidationMark's validateCell method to validate the cell event even when not triggering the columns event.

### Test Steps
1. Go to a page containing a Grid with dropdown dependency.
2. Change the parent cell value
3. Check that the child's OnCellValueChange is triggered just once for the child dropdown.
4. Press Ctrl+Z to undo
5. Check that the child's OnCellValueChange is triggered just once for the child dropdown.

### Screenshots
Before the changes:
dropdownDependencyIssue

After the changes: 
dropdownDependencyIssueSolved

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Columns/DropdownColumn.ts
+++ b/src/Providers/DataGrid/Wijmo/Columns/DropdownColumn.ts
@@ -160,13 +160,14 @@ namespace Providers.DataGrid.Wijmo.Column {
                     this.provider.index,
                     false
                 ) ?? '';
-            
+
             // if the child dropdown value is already empty, we don't want register the undo action and trigger cell value change event again
             if (column && currentValue !== '') {
                 // get original value of the child cell
                 const originalValue = this.grid.features.dirtyMark.getOldValue(
                     rowNumber,
-                    this.grid.getColumnByIndex(this.provider.index).config.binding
+                    this.grid.getColumnByIndex(this.provider.index).config
+                        .binding
                 );
 
                 const cellRange = new wijmo.grid.CellRange(
@@ -200,9 +201,7 @@ namespace Providers.DataGrid.Wijmo.Column {
                         ) ?? false;
 
                     // only add child undo action if it doesn't already exist. we don't want duplicated actions
-                    if (
-                        existingEditActionForColRow === false
-                    ) {
+                    if (existingEditActionForColRow === false) {
                         // add new child action into existing parent action in order
                         existingUndoAction.addChildAction(
                             new GridEditAction(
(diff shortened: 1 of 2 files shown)


bmarcelino-fe (approved):

src/Providers/DataGrid/Wijmo/Columns/DropdownColumn.ts
```diff
@@ -160,66 +160,80 @@ namespace Providers.DataGrid.Wijmo.Column {
...
+                const filteredUndoAction: wijmo.undo.UndoableAction[] =
+                    this.grid.features.undoStack.stack._stack.filter(
+                        (data: GridEditAction) =>
+                            data.col === currentSel.leftCol &&
+                            data.row === currentSel.topRow
                     );
+
+                // get the lastest undo action filtered if any
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                const existingUndoAction: any =
```
BenOsodrac, line 189: @OS-giulianasilva any way to avoid the any? I think in the past Gabriel did something about it, not sure what. 

I believe he started to create types somewhere, but I might be wrong.

BenOsodrac (approved):

src/Providers/DataGrid/Wijmo/Columns/DropdownColumn.ts
gnbm: @rugoncalves any insight?

OS-giulianasilva: @BenOsodrac I could not find this file that you mentioned. For now, I will proceed with the merge, but next week I can evaluate it with Rúben and correct it if necessary.

PR #381, opened 2023-11-06
ROU-4477: update version to 2.13.0

rugoncalves:
This PR is to update the version of ODC code to version 2.13.0


docs/assets/highlight.css
```diff
@@ -1,15 +1,19 @@
 :root {
-    --light-code-background: #FFFFFF;
-    --dark-code-background: #1E1E1E;
+    --light-code-background: #ffffff;
```
gnbm, line 2: @rugoncalves was this because of eslint/pretier? Or is there another reason?

gnbm (approved):

OS-giulianasilva (approved):

docs/assets/highlight.css
rugoncalves: I didn't change this manually. It's like this in the [main branch](https://github.com/OutSystems/outsystems-datagrid/blob/00d4fcf238c6897b1cabe3ee00ffc7667293298e/docs/assets/highlight.css#L2).

PR #382, opened 2023-11-20
ROU-3712: Dropdown cell change handler returns wrong Id if text is duplicate

OS-giulianasilva:
This PR is for fixing the dropdown with duplicated labels returning the wrong ID and not triggering OnCellValueChange

### What was happening
* When the Dropdown has duplicated labels with different values, the OnCellValueChange returns the wrong ID or is not triggered when changing to an option with the same label.

### What was done
* Now, we store the previous cell value into the beginningEdit, to validate if the value was changed into cellEditEnded. 

### Test Steps
1. Go to a page with a Grid containing dropdowns with duplicated option's labels
2. Change to another option with the same label
3. Check if the OnCellValue Changed was triggered and returns the correct value.

### Screenshots
Before:
dropdownDuplicatedLabelsIssue

After:
dropdownDuplicatedLabelsFixed

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
@@ -46,16 +46,6 @@ namespace Providers.DataGrid.Wijmo.Feature {
             // if the ESC key is pressed the e.cancel is true
             if (e.cancel) return;
 
-            // If the Delete or a BackSpace key is pressed, we assume that a change was made
-            if (
-                !!e.data &&
-                !!e.data.key &&
-                (e.data.key === 'Delete' || e.data.key === 'Backspace')
-            ) {
-                e.cancel = false;
-                return;
-            }
-
             // Stores the previous cell value to validate if it was changed in CellEditEnded handler.
             // Related to WJM-27988 that will be fixed in the Wijmo's 2023.2 release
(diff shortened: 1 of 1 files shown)


src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
```diff
@@ -38,61 +39,51 @@ namespace Providers.DataGrid.Wijmo.Feature {
          * For this purpose, here we consider null == undefined == ''
          * Bug: it does not work for checkboxes, since the activeEditor.value is always "on"
          */
-        private _cellEditBeforeEndingHandler(
+        private _beginningEditHandler(
             s: wijmo.grid.FlexGrid,
             e: wijmo.grid.CellRangeEventArgs
         ): void {
             // if the ESC key is pressed the e.cancel is true
             if (e.cancel) return;
```
rugoncalves, line 47: Maybe let's be consistent with how we test the `e.cancel`.

rugoncalves (approved):
Good job @OS-giulianasilva! Just pointed out a small improvement. But that's all

BenOsodrac (dismissed):

BenOsodrac (approved):

rugoncalves (approved):

PR #383, opened 2023-11-28
ROU-4486: changes required to expand the usage of setColumnFilterOptions method

rugoncalves:
This PR is not to be accepted!
It only contains the changes required to allow the action `API_Filter\SetColumnFilterOptions` to be used in more columns:
- Text
- Dropdown
- Checkbox 🌟
- Currency 🌟
- Date 🌟
- DateTime 🌟
- Number 🌟


OS-giulianasilva (approved):

OS-giulianasilva (changes requested):
@rugoncalves I've just remembered about the Error Message that needs to be changed now that this Client Action does not work only for Text and Dropdown.

JoaoFerreira-FrontEnd (approved):

gnbm:
@rugoncalves @OS-giulianasilva I assume we can close this PR since the feature was implemented in task ROU-4659?

PR #384, opened 2023-11-29
ROU-4651: solving incorrect outputs when API invokation fails

rugoncalves:
This PR contains a potential solution for how the APIs can start returning, in case of error, values more aligned with that.
The following APIs, were changed:
- `OutSystems.GridAPI.Rows.GetRowNumberByKey`
- `OutSystems.GridAPI.Pagination.GetCurrentPage`

image

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/OutSystems/GridAPI/Selection.ts
+++ b/src/OutSystems/GridAPI/Selection.ts
@@ -6,12 +6,12 @@ namespace OutSystems.GridAPI.Selection {
             gridID,
             errorCode:
                 OSFramework.DataGrid.Enum.ErrorCodes.API_FailedGetAllSelections,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getAllSelections();
-            }
+            },
+            hasValue: true
         });
 
         Performance.SetMark('Selection.GetAllSelections-end');
@@ -31,12 +31,12 @@ namespace OutSystems.GridAPI.Selection {
             errorCode:
                 OSFramework.DataGrid.Enum.ErrorCodes
                     .API_FailedGetAllSelectionsData,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getAllSelectionsData();
-            }
+            },
+            hasValue: true
         });
 
         Performance.SetMark('Selection.GetAllSelectionsData-end');
@@ -57,12 +57,12 @@ namespace OutSystems.GridAPI.Selection {
             errorCode:
                 OSFramework.DataGrid.Enum.ErrorCodes
                     .API_FailedGetCheckedRowsData,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getCheckedRowsData();
-            }
+            },
+            hasValue: true
         });
 
         Performance.SetMark('Selection.GetCheckedRowsData-end');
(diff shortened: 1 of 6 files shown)


src/OutSystems/GridAPI/Selection.ts
```diff
@@ -82,12 +82,12 @@ namespace OutSystems.GridAPI.Selection {
...
                 OSFramework.DataGrid.Enum.ErrorCodes
                     .API_FailedGetSelectedRowsCount,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getSelectedRowsCount();
-            }
+            },
+            hasValue: true
```
OS-giulianasilva, line 90: I think here we can add `defaultFailValue: -1` too

src/OutSystems/GridAPI/Selection.ts
```diff
@@ -132,12 +132,12 @@ namespace OutSystems.GridAPI.Selection {
...
                 OSFramework.DataGrid.Enum.ErrorCodes
                     .API_FailedGetSelectionAverage,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getSelectionAverage();
-            }
+            },
+            hasValue: true
```
OS-giulianasilva, line 140: Same here

src/OutSystems/GridAPI/Selection.ts
```diff
@@ -181,12 +182,12 @@ namespace OutSystems.GridAPI.Selection {
...
             errorCode:
                 OSFramework.DataGrid.Enum.ErrorCodes.API_FailedGetSelectionMax,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getSelectionMaxMin(true);
-            }
+            },
+            hasValue: true
```
OS-giulianasilva, line 190: Not sure about here, since the max and min values can be -1 and valid, but maybe we can add for consistency,

src/OutSystems/GridAPI/Selection.ts
```diff
@@ -205,12 +206,12 @@ namespace OutSystems.GridAPI.Selection {
...
             errorCode:
                 OSFramework.DataGrid.Enum.ErrorCodes.API_FailedGetSelectionMin,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getSelectionMaxMin(false);
-            }
+            },
+            hasValue: true
```
OS-giulianasilva, line 214: Same as GetSelectionMAx

src/OutSystems/GridAPI/Selection.ts
```diff
@@ -229,12 +230,12 @@ namespace OutSystems.GridAPI.Selection {
...
             errorCode:
                 OSFramework.DataGrid.Enum.ErrorCodes.API_FailedGetSelectionSum,
-            hasValue: true,
             callback: () => {
                 const grid = GridManager.GetGridById(gridID);
 
                 return grid.features.selection.getSelectionSum();
-            }
+            },
+            hasValue: true
```
OS-giulianasilva, line 238: Same as GetSelectionMAx

src/OutSystems/GridAPI/Selection.ts
rugoncalves: This one, I'm not sure, as -1 is a valid value... better would be not to return any value at all... but not possible... as such left 0 as an error value - discuss-able.

src/OutSystems/GridAPI/Selection.ts
rugoncalves: Same as here as the previous comment: _This one, I'm not sure, as -1 is a valid value... better would be not to return any value at all... but not possible... as such left 0 as an error value - discuss-able._

src/OutSystems/GridAPI/Selection.ts
rugoncalves: Same as here as GetSelectionMAx :)

src/OutSystems/GridAPI/Selection.ts
rugoncalves: Same as here as GetSelectionMAx :)

OS-giulianasilva (approved):

joselrio (approved):

PR #385, opened 2023-12-13
ROU-4407: Open link with middle button on action column

bmarcelino-fe:
This PR is for allow using the middle mouse button to open link on the action columns

### What was done
* Added a validation to element type link, to apply different attributes based on externalURL to allows to redirect to an external URL on an action column, using a URL from data or a static URL. If the ExternalURL are being defined, the click event will not be triggered.

### Test Steps
**Test case 1**
1. Open sample page
2. Open each link with middle mouse button
3. Expected: Each link will open in a separated page

**Test Case 2**
1. Open sample page
2. On last column, click on link
3. Expected: The click event will be triggered and a feedback message will appear

### Checklist
* [X] tested locally
* [X] documented the code

### DIFF
--- a/src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
+++ b/src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
@@ -23,6 +23,7 @@ namespace OSFramework.DataGrid.Configuration.Column {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         public editor: any;
         public errorMessage: string;
+        public externalLink: string;
         public format: string;
         public genericColumnId: string;
         public header: string;
--- a/src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
+++ b/src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
@@ -23,7 +23,6 @@ namespace OSFramework.DataGrid.Configuration.Column {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         public editor: any;
         public errorMessage: string;
-        public externalURL: string;
         public format: string;
         public genericColumnId: string;
         public header: string;
--- a/src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
+++ b/src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
@@ -23,7 +23,7 @@ namespace OSFramework.DataGrid.Configuration.Column {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         public editor: any;
         public errorMessage: string;
-        public externalLink: string;
+        public externalURL: string;
         public format: string;
         public genericColumnId: string;
         public header: string;
--- a/src/OSFramework/DataGrid/Types/index.ts
+++ b/src/OSFramework/DataGrid/Types/index.ts
@@ -137,6 +137,7 @@ namespace OSFramework.DataGrid.Types {
         canBeHidden: boolean;
         conditionalFormat: Array<OSStructure.ConditionalFormat>;
         errorMessage: string;
+        externalLink: string;
         genericColumnId: string;
         headerTooltip: string;
         isMandatory: boolean;
@@ -152,6 +153,7 @@ namespace OSFramework.DataGrid.Types {
 
     export interface IActionColumnExtraConfigs extends IColumnExtraConfigs {
         actionColumnElementType: DataGrid.Enum.CellTemplateElementType;
+        externalLink?: string;
     }
 
     export interface IImageColumnExtraConfigs
--- a/src/OSFramework/DataGrid/Types/index.ts
+++ b/src/OSFramework/DataGrid/Types/index.ts
@@ -137,7 +137,7 @@ namespace OSFramework.DataGrid.Types {
         canBeHidden: boolean;
         conditionalFormat: Array<OSStructure.ConditionalFormat>;
         errorMessage: string;
-        externalLink: string;
+        externalURL: string;
         genericColumnId: string;
         headerTooltip: string;
         isMandatory: boolean;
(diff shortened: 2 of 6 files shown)


src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
```diff
@@ -23,6 +23,7 @@ namespace OSFramework.DataGrid.Configuration.Column {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         public editor: any;
         public errorMessage: string;
+        public externalURL: string;
```
OS-giulianasilva, line 26: I think we can remove this parameter from here and only maintain on the ColumnConfigAction Class only since this is not common to all columns.

src/OSFramework/DataGrid/Types/index.ts
```diff
@@ -137,6 +137,7 @@ namespace OSFramework.DataGrid.Types {
         canBeHidden: boolean;
         conditionalFormat: Array<OSStructure.ConditionalFormat>;
         errorMessage: string;
+        externalURL: string;
```
OS-giulianasilva, line 140: Same thing here.

src/Providers/DataGrid/Wijmo/Columns/ImageColumn.ts
```diff
@@ -39,6 +39,7 @@ namespace Providers.DataGrid.Wijmo.Column {
             config.cellTemplate = Helper.CellTemplateFactory.MakeCellTemplate(
                 this.config.actionColumnElementType,
                 config.binding,
+                this.config.externalURL,
```
OS-giulianasilva, line 42: We've discussed it before, but now I think this parameter should be optional in MakeCellTemplate function, and also placed after the altText parameter. This way we don't need to add it here, since it does not make sense in the ImageColumn.

src/Providers/DataGrid/Wijmo/Helper/CellTemplateFactory.ts
```diff
@@ -9,12 +9,17 @@ namespace Providers.DataGrid.Wijmo.Helper.CellTemplateFactory {
     export function MakeCellTemplate(
         type: OSFramework.DataGrid.Enum.CellTemplateElementType,
         binding: string,
+        externalURL: string,
```
OS-giulianasilva, line 12: As mentioned before, I think this should be the last optional parameter here.

OS-giulianasilva (changes requested):

src/Providers/DataGrid/Wijmo/Helper/CellTemplateFactory.ts
bmarcelino-fe: Done

src/Providers/DataGrid/Wijmo/Columns/ImageColumn.ts
bmarcelino-fe: Done

src/OSFramework/DataGrid/Types/index.ts
bmarcelino-fe: Done

src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
bmarcelino-fe: Done

src/Providers/DataGrid/Wijmo/Helper/CellTemplateFactory.ts
```diff
@@ -45,12 +50,34 @@ namespace Providers.DataGrid.Wijmo.Helper.CellTemplateFactory {
                 });
                 break;
             case OSFramework.DataGrid.Enum.CellTemplateElementType.Link: {
-                cellTemplate = wijmo.grid.cellmaker.CellMaker.makeLink({
+                console.log('here', externalURL);
```
OS-giulianasilva, line 53: Please remove this console.log

OS-giulianasilva (approved):

src/Providers/DataGrid/Wijmo/Helper/CellTemplateFactory.ts
bmarcelino-fe: done

OS-giulianasilva (approved):

src/Providers/DataGrid/Wijmo/Helper/CellTemplateFactory.ts
```diff
@@ -10,11 +10,16 @@ namespace Providers.DataGrid.Wijmo.Helper.CellTemplateFactory {
...
         binding: string,
         callback: (item) => void,
-        altText?: string
+        altText?: string,
+        externalURL?: string
     ): wijmo.grid.ICellTemplateFunction {
         let cellTemplate: wijmo.grid.ICellTemplateFunction;
 
         const hasFixedText = binding.startsWith('$');
+        const hasExternalURL = externalURL?.substring(0, 4) === 'http';
```
JoaoFerreira-FrontEnd, line 19: I don't know if this validation is enough, since nowadays we can use links without the http or https or even without the www... Why we didn't just check if externalURL it's filled?

src/Providers/DataGrid/Wijmo/Helper/CellTemplateFactory.ts
```diff
@@ -45,12 +50,32 @@ namespace Providers.DataGrid.Wijmo.Helper.CellTemplateFactory {
...
+                const cellTemplateOptions = {
                     text,
-                    click: (e, ctx) => {
+                    href: undefined,
+                    attributes: undefined,
+                    click: undefined
+                };
+
+                // Validate if is a link and defined the default options
+                if (externalURL === '') {
```
JoaoFerreira-FrontEnd, line 62: like we are doing here

JoaoFerreira-FrontEnd (changes requested):

JoaoFerreira-FrontEnd (approved):

PR #392, opened 2024-01-17
ROU-4691: Column Filter does not reflect the correct value when using FilterByCondition API

OS-giulianasilva:
This PR is to fix the filter popup not displaying the correct value when using FilterByCondition API in a Currency Column. It also corrects the issue of the filter not being applied to the correct column when the Grid has columns with the same binding.

### What was happening
* When using the FilterByCondition API to set a filter on a Currency column, the value passed to the _getFilterConditionValue was a string instead of a number.
   * This prevented the condition value from appearing correctly in the filter popup. 
* When using the FilterByCondition API to set a filter to a column with duplicated binding in a Grid, the filter was always applied to the first column corresponding to the binding.

### What was done
* In the ColumnFilter.byCondition method, added an evaluation was added to check if the target column is numerical. If true, we check if the value is a number and convert it.
* In the ColumnFilter.byCondition and ColumnFilter.byValue, edit how a filter is applied to a column. Instead of using the binding, now we use the column provider's instance.

#### Test 1:
1. Go to a page with a Grid that contains a Currency column
2. Use the FilterByCondition API to set a new filter to the Currency column
3. Open the column filter
4. Check that the condition value is displayed correctly

#### Test 2:
1. Go to a page with a Grid that contains two columns with the same binding
2. Use the FilterByCondition API to set a new filter to a column
3. Check that the desired column is filtered correctly
4. Repeat for another column with the same binding.

### Screenshots
Before first fix: 
filterbycondition_isse

After first fix:
filterbycondition_fix

Before second fix:
filterbycondition_samebinding_issue

After second fix:
filterbycondition_samebinding_fix

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/OSFramework/DataGrid/Enum/ErrorMessages.ts
+++ b/src/OSFramework/DataGrid/Enum/ErrorMessages.ts
@@ -14,6 +14,7 @@ namespace OSFramework.DataGrid.Enum {
         AutoGenerateGridWrite = 'The auto generation of columns based on Data is not available when the Grid is editable. Please use ArrangeData action for this scenario.',
         Column_NotFound = 'Column not found',
         CustomizeExportingMessageEmptyString = 'It seems you are passing an empty message. Please pass a valid message.',
+        Filter_InvalidDataType = 'It seems you are trying to filter a Number/Currency column with a value that is not a number.',
         FreezeColumnPositiveNumberExpected = 'Unable to freeze column. Please use a positive number.',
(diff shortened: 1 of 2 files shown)


bmarcelino-fe:
@OS-giulianasilva please check the Sonar Cloud because it failed on Quality Gate.

bmarcelino-fe (approved):

BenOsodrac (approved):

PR #402, opened 2024-03-19
ROU-4768: Fix OnCellValueChange returning the dropdown label instead of the value

OS-giulianasilva:
This PR is for fixing the OnCellValueChange returning the dropdown label instead of the value when the SetCellData is called.

### What was happening
* When triggering the OnCellValueChange after calling the SEeCellData, it returned the dropdown's formatted value (label) instead of the option id/key. 

### What was done
* In the Wijmo's getCellData method, we now set the last argument to false, this way we get the dropdown option id/key instead of the dropdown label.

### Checklist
* [x] tested locally
* [x] documented the code
* [] requires changes in OutSystems (if so, provide a module with changes)

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
@@ -602,11 +602,7 @@ namespace Providers.DataGrid.Wijmo.Feature {
 			triggerOnCellValueChange = true
 		): void {
 			// This method gets executed by an API. No values change in columns, so the current value and the original one (old value) are the same.
-			const currValue = this._grid.provider.getCellData(
-				rowNumber,
-				column.provider.index,
-				column.columnType === OSFramework.DataGrid.Enum.ColumnType.Dropdown
-			);
+			const currValue = this._grid.provider.getCellData(rowNumber, column.provider.index, false);
 
 			//If we decide not to trigger the column events we will skip this step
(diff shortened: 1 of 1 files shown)


joselrio (approved):

rugoncalves (changes requested):
These changes leave me a bit uncomfortable. The reason is that in the case of the dropdown, when the `getCellData` receives as the 3rd parameter:
   - true: it will return the label (e.g.: "_Greece_")
   - false: it will return the value (e.g.: "_5_")

So this even appears to cause a breaking change, when triggering the event... Have all situations and circumstances being analyzed?

OS-giulianasilva:
> These changes leave me a bit uncomfortable. The reason is that in the case of the dropdown, when the `getCellData` receives as the 3rd parameter:
> 
> * true: it will return the label (e.g.: "_Greece_")
> * false: it will return the value (e.g.: "_5_")
> 
> So this even appears to cause a breaking change, when triggering the event... Have all situations and circumstances been analyzed?

@rugoncalves When a dropdown is manually edited, we return the dropdown value instead of the label in the OnCellValueChange event. When the SetCellData is used, we return the label instead. So the purpose is to keep it consistent with the default behavior. 
Additionally, I performed some manual tests and ran the pipeline, and no issues were found.

rugoncalves (approved):
Consider adding additional comments in the code.

PR #425, opened 2024-07-26
ROU-10879: CTRL+SHIFT+A Removes selected mandatory column

OS-giulianasilva:
This PR is to fix the issue when performing Ctrl+A inside a column picker

### What was happening
* When a column has the `CanBeHidden` property set to false, its checkbox inside the Column picker became disabled
* However, when performing `Ctrl+A` inside the Column picker several times, it is possible to change its visibility.

### What was done
* Added a event listener to the column picker HTML element as workaround
    * For all the disabled items in the column picker:
        * Checks if the column must be visible and if it is already in the checkedColumns array. If not, we add it.
        * Checks if the column must NOT be visible and if it is in the checkedColumns array. If yes, we remove it.
* This workaround does not work for the case where the `CanBeHidden = true` and `Visible = false` 

### Test Steps
1. Go to a test page with a Grid
2. Set one column with the CanBeHidden property to false
3. Open the Grid's Column picker
4. Perform Ctrl+A multiple times
5. Check that the column is not hidden 

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
@@ -116,6 +116,7 @@ namespace Providers.DataGrid.Wijmo.Feature {
 		private _configureCheckbox(col: OSFramework.DataGrid.Column.IColumn, item: HTMLElement) {
 			if (col.config.canBeHidden === false) {
 				const checkbox = item.querySelector('input[type="checkbox"]');
+				wijmo.toggleClass(item, 'wj-state-disabled', true);
 				if (checkbox) {
 					checkbox.setAttribute('disabled', 'true');
 				}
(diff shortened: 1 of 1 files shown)


src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
```diff
@@ -232,6 +266,9 @@ namespace Providers.DataGrid.Wijmo.Feature {
 
 				e.preventDefault();
 			};
+
+			// This code fixes WJM-34234
+			this._theColumnPicker.hostElement.addEventListener('keydown', this._issueWorkaround.bind(this));
```
BenOsodrac, line 271: don't we need to remove this event somewhere?

OS-giulianasilva: This was something that I didn't pay attention before, but I just found out the this._theColumnPicker.dispose() already takes care of it according to the [documentation](https://developer.mescius.com/wijmo/api/classes/Wijmo_Input.Listbox.html#dispose)

BenOsodrac (approved):

src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
```diff
@@ -116,6 +116,7 @@ namespace Providers.DataGrid.Wijmo.Feature {
 		private _configureCheckbox(col: OSFramework.DataGrid.Column.IColumn, item: HTMLElement) {
 			if (col.config.canBeHidden === false) {
 				const checkbox = item.querySelector('input[type="checkbox"]');
+				wijmo.toggleClass(item, 'wj-state-disabled', true);
```
rugoncalves, line 119: Will this be also removed with the workaround?

src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
```diff
@@ -168,6 +169,39 @@ namespace Providers.DataGrid.Wijmo.Feature {
 			);
 		}
 
+		// This code fixes WJM-34234
+		private _issueWorkaround(e) {
```
rugoncalves, line 173: missing return type

src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
```diff
@@ -168,6 +169,39 @@ namespace Providers.DataGrid.Wijmo.Feature {
...
+			if (e.ctrlKey && e.code === 'KeyA') {
+				const checkedItems = [...this._theColumnPicker.checkedItems];
+				this._theColumnPicker._children.forEach((item, index) => {
+					const input = item.querySelector('input');
+
+					// If the checkbox item is disabled
+					if ((input && input.disabled) || item.classList.contains('wj-state-disabled')) {
```
rugoncalves, line 180: Do we need to compare with `input && input.disabled`, as in, is it not enough to use the second part of the condition (`item.classList.contains('wj-state-disabled')`)?

OS-giulianasilva: I understand your point, but this is what we have in the workaround provided, so I just assumed that it should be like this.

src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
OS-giulianasilva: Done

src/Providers/DataGrid/Wijmo/Features/ColumnPicker.ts
rugoncalves: Got it, thank you!

rugoncalves (approved):

BenOsodrac (approved):

PR #426, opened 2024-07-26
ROU-4896: Remove OnCellValueChange trigger from ApplyRowValidations API

OS-giulianasilva:
This PR is to remove the OnCellValueChange trigger from ApplyRowValidations API

### What was happening
- When using the ApplyRowValidations Client Action in a Grid with Dropdown Dependency, the child dropdown cell was being erased
- This happened because the ApplyRowValidations triggers the OnCellValueChange event of the Column and the Dropdown Column's _parentCellValueChangeHandler method was not considering this scenario.

### What was done
* Removed the OnCellValueChange event trigger when calling the ApplyRowValidations. 
* This intruduces a breaking change.

### Test Steps

1. Go to a page with a Grid with Dropdown Dependency
2. Run the ApplyRowValidations Client Action
3. Check that no cells were erased
4. Check that the OnCellValueChange was not triggered

### Checklist
* [x] tested locally
* [x] documented the code


BenOsodrac (approved):

rugoncalves (approved):
This is a breaking of behavior, and needs to be well noted and documented in the release notes.

PR #427, opened 2024-07-26
ROU-10774: Fix additional spaces when adding a new class to cssClass and cssClassAll

OS-giulianasilva:
This PR is to fix the additional spaces when adding a new class to cssClass and cssClassAll.

### What was done
* Check if classList is not empty before splitting the string into an array. If empty, return a empty array.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/Styling.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/Styling.ts
@@ -13,8 +13,8 @@ namespace Providers.DataGrid.Wijmo.Feature {
 			let classList = [];
 			const column = this._grid.getColumn(columnID);
 
-			if (column) {
-				const classListString = column.provider.cssClass || ' ';
+			if (column && column.provider.cssClass) {
+				const classListString = column.provider.cssClass;
 				classList = classListString.split(' ');
 			}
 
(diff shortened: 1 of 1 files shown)


src/Providers/DataGrid/Wijmo/Features/Styling.ts
```diff
@@ -25,8 +25,8 @@ namespace Providers.DataGrid.Wijmo.Feature {
 			let classList = [];
 			const column = this._grid.getColumn(columnID);
 
-			if (column) {
-				const classListString = column.provider.cssClassAll || ' ';
+			if (column && column.provider.cssClassAll) {
+				const classListString = column.provider.cssClassAll;
```
rugoncalves, line 29: This const could be created before the `if`, no?

rugoncalves (changes requested):

src/Providers/DataGrid/Wijmo/Features/Styling.ts
OS-giulianasilva: Well, it could, but it isused inside the if only, so I don't see why.

rugoncalves: not really, since you're using it in the condition! :P

rugoncalves (approved):

src/Providers/DataGrid/Wijmo/Features/Styling.ts
rugoncalves: not very relevant

BenOsodrac (approved):

PR #431, opened 2024-08-20
ROU-11001: Added sanitizeInputValues configuration

gnbm:
This PR is for adding the new `sanitizeInputValues` configuration.

### What was happening
* When passing HTML to the Grid, mainly on Action and Image Columns and the ContextMenu., we opened a door to an XSS vulnerability.

### What was done
* Added a new `sanitizeInputValues` configuration to allow the developer to control when to sanitize the Grid's data.
* The default will be `sanitizeInputValues = true` so this will be a needed **breaking change**.

### Test Steps
1. Open a screen with a Link Column 
2. On the Get From Other Sources add to the first position the following text `Test `
3. Check that now we can see the text and no code is executed

### Screenshots
- Before:

image

- After the fix:

image

### Checklist
* [X] tested locally
* [X] documented the code
* [X] requires changes in OutSystems 
* [ ] requires new sample page in OutSystems

### DIFF
--- a/src/OutSystems/GridAPI/GridManager.ts
+++ b/src/OutSystems/GridAPI/GridManager.ts
@@ -15,6 +15,7 @@ namespace OutSystems.GridAPI.GridManager {
 			let output = false;
 			if (grid !== undefined) {
 				if (grid.isReady && data !== '' && data !== '{}') {
+					if (grid.config.sanitizeInputValues) data = OSFramework.DataGrid.Helper.Sanitize(data);
 					grid.setData(data);
 				}
 				output = true;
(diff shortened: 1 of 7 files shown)


src/OutSystems/GridAPI/GridManager.ts
```diff
@@ -15,6 +15,8 @@ namespace OutSystems.GridAPI.GridManager {
 			let output = false;
 			if (grid !== undefined) {
 				if (grid.isReady && data !== '' && data !== '{}') {
+					// When the configurantion is set to sanitize the input values, we need to sanitize the data before setting it
+					if (grid.config.sanitizeInputValues) data = OSFramework.DataGrid.Helper.Sanitize(data);
```
rugoncalves, line 19: Add curly brackets, to make the code more readable.

rugoncalves (changes requested):

src/OutSystems/GridAPI/GridManager.ts
gnbm: Done @rugoncalves

rugoncalves (dismissed):

rugoncalves (approved):

rugoncalves (approved):
Looking good!

OS-giulianasilva (approved):

PR #437, opened 2024-10-29
ROU-11327: Fix tooltip issue with Grouped Columns and increment version

gnbm:
### What was happening
* There was an issue where the column headers displayed incorrect tooltip content when using grouped columns in a Grid:

### What was done
* Improved the validation about the tooltip method to be called on the internal event handler `_onMouseEnter(e)` because if the _Column Header_ is from a _Group Column_, we need to use a different approach than the regular header.
* We can check if the current target is a _ColumnGroup_ by checking its CSS class `wj-colgroup` and, if there is no binding and the number of `columns` inside _ColumnGroup_ > 0 or if the number of `columnGroups` inside _ColumnGroup_ > 0.

### Test Steps
1. Go to a screen with two Grids, one with Grouped Columns and another without
2. On the one with Grouped Columns hover all Column Headers and check that the tooltip content is the expected
3. Repeat the same steps for the one without Grouped Columns for regression checks

### Screenshots

- **Issue**:

ToolttipGroupHeaderIssues

- **Fixed Issue**:

TooltipGroupedColumnsIssueFixed

### Checklist
* [X] tested locally
* [X] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ToolTip.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ToolTip.ts
@@ -33,12 +33,8 @@ namespace Providers.DataGrid.Wijmo.Feature {
 				}
 			} else if (cellType === wijmo.grid.CellType.ColumnHeader) {
 				// If the Column Header is from a Group Column, we need to use a different approach than the regular header
-				// We can check if the current target is a ColumnGroup by checking its class and if providerIndex is -1
-				if (
-					_currTarget.classList.contains(Helper.Constants.CssClasses.ColumnGroup) &&
-					this._grid.getColumnByIndex(ht.getColumn().index).providerIndex === -1
-				) {
-					this._setColumnGroupHeaderTooltip(_currTarget);
+				if (_currTarget.classList.contains(Helper.Constants.CssClasses.ColumnGroup)) {
+					this._setColumnGroupHeaderTooltip(_currTarget, ht);
 				} else {
 					this._setHeaderTooltip(_currTarget, ht);
 				}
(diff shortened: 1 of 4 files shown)


src/Providers/DataGrid/Wijmo/Features/ToolTip.ts
```diff
@@ -32,8 +32,16 @@ namespace Providers.DataGrid.Wijmo.Feature {
...
-				if (_currTarget.classList.contains(Helper.Constants.CssClasses.ColumnGroup)) {
+				// If the Column Header is from a Group Column, we need to use a different approach than the regular header
+				// We can check if the current target is a ColumnGroup by checking its class and if it has columns or columnGroups
+				if (
+					_currTarget.classList.contains(Helper.Constants.CssClasses.ColumnGroup) &&
+					((ht.getColumn(true)._binding === undefined &&
+						(ht.getColumn(true) as wijmo.grid.ColumnGroup).columns !== undefined &&
+						(ht.getColumn(true) as wijmo.grid.ColumnGroup).columns.length > 0) ||
+						((ht.getColumn(true) as wijmo.grid.ColumnGroup).columnGroups !== undefined &&
+							(ht.getColumn(true) as wijmo.grid.ColumnGroup).columnGroups.length > 0))
```
BenOsodrac, line 43: Maybe `ht.getColumn(true)` could be stored on a variable before the if. This way it would make the code cleaner and reduce the method calls to only one?

gnbm: Makes total sence @BenOsodrac ! Changes done

BenOsodrac (approved):

OS-giulianasilva (approved):

PR #438, opened 2024-11-07
ROU-11338: Dropdown workaround arrows

rugoncalves:
This PR for adding the workaround provided by wijmo, to fix the issue of when using the dropdown with the keyboard the values that are displayed are `[object Object]`. This issue is associated with the RPM-5312.

### What was happening
* When using the keyboard arrows, the dropdown would not display the actually selected element (text) but rather `[object Object]`

### What was done
* Added the workaround provided by wijmo together with the comments

### Test Steps
1. Go to the test page
2. Open a dropdown
3. Move between elements in the dropdown using the arrow keys
4. The dropdown should display that are currently elected

### Screenshots
- before
dropdown-arrows-issues
- after
dropdown-arrows-issues-fixed

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Grid/FlexGrid.ts
+++ b/src/Providers/DataGrid/Wijmo/Grid/FlexGrid.ts
@@ -8,8 +8,15 @@ namespace Providers.DataGrid.Wijmo.Grid {
 		implements IGridWijmo
 	{
 		private _fBuilder: Feature.FeatureBuilder;
+
+		//TODO: Workaround to be removed with wijmo update (2015v1?). See task ROU-11338.
+		private _inUseWorkaroundRou11338: boolean;
+
 		private _resizedColumnHandler: OSFramework.DataGrid.Callbacks.Generic;
 		private _rowMetadata: RowMetadata;
+
+		//TODO: Workaround to be removed with wijmo update (2015v1?). See task ROU-11338.
+		private _useWorkaroundRou11338: boolean;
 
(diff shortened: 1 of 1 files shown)


src/Providers/DataGrid/Wijmo/Grid/FlexGrid.ts
```diff
@@ -96,7 +168,14 @@ namespace Providers.DataGrid.Wijmo.Grid {
 		public addColumn(col: OSFramework.DataGrid.Column.IColumn): Promise<void> {
 			super.addColumn(col);
 
+			//TODO: Workaround to be removed with wijmo update (2015v1?). See task ROU-11338.
+			this._useWorkaroundRou11338 =
+				this._useWorkaroundRou11338 || col.columnType === OSFramework.DataGrid.Enum.ColumnType.Dropdown;
+
 			if (this.isReady) {
+				//TODO: Workaround to be removed with wijmo update (2015v1?). See task ROU-11338.
+				this._dropdownWorkaround();
```
OS-giulianasilva, line 177: Why is it necessary to call the `this._dropdownWorkaround` here and during the build?

rugoncalves: There are 2 situations:
- when the grid is built having a dropdown column already
- when the grid is built without a dropdown column, and a dropdown column is added afterwards

The idea of this approach is to avoid adding the handlers when no dropdown column is used.

OS-giulianasilva (approved):

PR #444, opened 2025-02-05
ROU-11578: Some HTML text is kept encoded when exported into CSV

OS-giulianasilva:
This PR is for fix encoded symbols and HTML tags that are presented in the exported CSV when a Grid has columns in the Group Panel.

### What was happening
* When a Grid had columns in the Group Panel and and CSV file was exported, the generated files was created with encoded symbols and HTML text;
* Example:
    * Given the text: **T&st N%m& ‹›**
    * The exported text would be: \<b\>T&amp\;st N%m&amp\; ‹›\<\/b\>
* The bold tags are added by default when a column is grouped. Check the example below that shows the `Product Name` Column grouped:
image

### What was done
* Added a new `_stripHtmlTags` method to the `GroupPanel` feature to be used for removing HTML tags before exporting the CSV.
* Set the `isContentHtml` property to `true` for every column that is added to the Group Panel.
    * This prevents the symbols to be encoded when exported to CSV.
    * When the column is removed from the Group Panel, we set the property to `false`.

### Test Steps
1. Go to a page that contains a Grid with data with symbols.
2. Add one column to the Group Panel.
3. Export the Grid's CSV.
4. Check that the grouped column data does not contain HTML tags or encoded symbols.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/GroupPanel.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/GroupPanel.ts
@@ -38,7 +38,7 @@ namespace Providers.DataGrid.Wijmo.Feature {
 			OSFramework.DataGrid.Interface.IBuilder,
 			OSFramework.DataGrid.Interface.IDisposable
 	{
-		private _currGroupDescription: Array<wijmo.collections.PropertyGroupDescription>;
+		private _currGroupDescription = new Array<wijmo.collections.PropertyGroupDescription>();
 		private _grid: Grid.IGridWijmo;
 		private _groupPanel: wijmo.grid.grouppanel.GroupPanel;
 		private _panelId: string;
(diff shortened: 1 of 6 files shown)


joselrio (approved):

src/Providers/DataGrid/Wijmo/Features/GroupPanel.ts
```diff
@@ -38,7 +38,7 @@ namespace Providers.DataGrid.Wijmo.Feature {
 			OSFramework.DataGrid.Interface.IBuilder,
 			OSFramework.DataGrid.Interface.IDisposable
 	{
-		private _currGroupDescription: Array<wijmo.collections.PropertyGroupDescription>;
+		private _currGroupDescription = new Array<wijmo.collections.PropertyGroupDescription>();
```
rugoncalves, line 41: Why are we now instantiating directly the array?

src/Providers/DataGrid/Wijmo/Features/GroupPanel.ts
```diff
@@ -104,11 +104,33 @@ namespace Providers.DataGrid.Wijmo.Feature {
...
+						}
+					});
+
+					// Loop through the group descriptions just removed from the Group Panel and set isContentHtml to false.
+					ungroupedDescriptions.forEach(function (groupDesc: wijmo.collections.PropertyGroupDescription) {
+						const col = grid.provider.getColumn(groupDesc.propertyName);
+						if (col) {
+							col.isContentHtml = false;
+						}
+					});
```
rugoncalves, line 133: Wer're doing basically, 4 iteractions... can we make simply 2?

rugoncalves (changes requested):

src/Providers/DataGrid/Wijmo/Features/GroupPanel.ts
OS-giulianasilva: because of the new code where we iterate over it (using `filter` and `includes`)

rugoncalves (dismissed):
Unblocking the PR.

src/Providers/DataGrid/Wijmo/Features/GroupPanel.ts
OS-giulianasilva: you're correct!

rugoncalves (approved):

PR #458, opened 2025-07-09
ROU-12064: Pasting into Grid cell followed by Right-Click copy marks cell as unedited

OS-giulianasilva:
This PR is for fix issue when a cell is copied using the Context Menu after being edited.

### What was happening
* When a cell was edited and then copied using the Context Menu, the change was not being tracked when running the Get Changed Lines.

### What was done
* Now we set the focus to the active cell after finishing the Context Menu action.
    * This is is similar what is implemented in [this Wijmo example](https://developer.mescius.com/wijmo/demos/Grid/ContextMenus/purejs).
image
    * Despite in the example the Grid is refreshed, this was not suitable for other use cases. For example, when adding a new row via Context Menu, the dirty marks were not appearing because the Grid has been refreshed.

### Test Steps
1. Go to a test page.
2. Edit a cell.
3. In the same cell, right click to open the Context Menu.
4. Click "Copy".
5. Run the Get Changed Lines Action.
6. Check that the change was tracked correctly.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ContextMenu.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ContextMenu.ts
@@ -262,9 +262,18 @@ namespace Providers.DataGrid.Wijmo.Feature {
 		 */
 		private _raiseClickEvent(menuItemId: string): void {
 			const menuItem = this._menuItems.get(menuItemId);
+			const provider = this.grid.provider;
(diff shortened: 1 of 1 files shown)


joselrio (approved):

gnbm (approved):
LGTM

PR #479, opened 2025-12-15
ROU-12281: Fix dirty mark issue with undo after Mark Changes As Saved

OS-giulianasilva:
This pull request fixes an issue that caused the dirty mark to not be added when performing undo after marking changes as saved.  It improves the undo functionality in the Wijmo DataGrid provider by ensuring that original cell values are properly saved before an undo action is performed. The most significant changes are as follows:

**Undo stack event handling:**

* Added a private `_undoingActionHandler` method to `UndoStack.ts` that listens for undo actions. When a grid edit action is being undone, it saves the original value of the affected cell using the `dirtyMark` feature.
* Registered the `_undoingActionHandler` as an event handler for the `undoingAction` event on the `wijmo.undo.UndoStack` instance, ensuring the handler is called whenever an undo is triggered.

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/UndoStack.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/UndoStack.ts
@@ -20,6 +20,15 @@ namespace Providers.DataGrid.Wijmo.Feature {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


JoaoFerreira-FrontEnd (approved):
LGTM

PR #489, opened 2026-02-19
ROU-11908: Fixed multiple requests for language files

os-davidlourenco:
This PR is to fix the multiple requests sent for language files when multiple Grids are displayed on the same page

### What was happening
* When a screen had multiple Grids, multiple requests for the language files were being made

Screen Recording 2026-02-19 at 18 28 09

### What was done
* Logic was changed to ensure that only one request is made per page, even though multiple Grids exist

### Test Steps
1. Open the browser with developer tools in the network tab
2. Open the sample application 
3. Filter by 'wijmo_culture' file name
4. Validate that only one request is done

### Screenshots
Screen Recording 2026-02-19 at 18 27 40

### Checklist
* [x] tested locally
* [x] documented the code


rugoncalves (approved):
Looking good!

PR #500, opened 2026-04-14
ROU-12718: Update lodash version

gnbm:
### What was done
* This pull request makes a minor update to the `@types/lodash` development dependency in the `package.json` file, upgrading it from version 4.17.23 to 4.17.24.
* This was required to fix [CVE-2025-13465](https://nvd.nist.gov/vuln/detail/CVE-2025-13465).

### Test Steps
- Ran tests pipeline

### Checklist
* [x] tested locally
* [ ] documented the code

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "gulp-remove-empty-lines": "^0.1.0",
     "gulp-sourcemaps": "^3.0.0",
     "gulp-typescript": "^6.0.0-alpha.1",
+    "lodash": "4.17.23",
     "postcss": "^8.5.6",
     "postcss-discard-comments": "^5.1.2",
     "postcss-discard-duplicates": "^5.1.0",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "UI Components"
   ],
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "UI Components"
   ],
   "devDependencies": {
-    "@types/lodash": "^4.17.23",
+    "@types/lodash": "^4.17.24",
```
rugoncalves, line 33: Do we need `@types/lodash`? Besides that the types and the version of `lodash` don't match...

rugoncalves (commented):
Have a question ☝️

package.json
gnbm: @rugoncalves types were already here, not matching versions, and there's no 1-to-1 match.
What I did was check the lodash version that was being installed in `node_modules` after running `npm i`

rugoncalves: The question is, do we need the types given that we now added the package as devDependency?

rugoncalves: As we saw, we do need the `@types`.

rugoncalves (approved):
Looking good!

PR #507, opened 2026-06-02
ROU-12796: Fix seconds trim on SetCellData

os-davidlourenco:
This PR is to fix the seconds trim on SetCellData

### What was happening
* When using the API method `SetCellData` if the column was of type DateTime the seconds part was trimmed due to the fact that the inline picker only allows the selection of hours and minutes.
* If the user defines the format as containing seconds, the seconds would always be 0. This was inconsistent with the insertion of data by the DataAction that would include the seconds.

### What was done
* The trim method was removed as well as the call to the method
* The cell equals function was updated to take into consideration the possibility of having dates with and without seconds.

### Test Steps
1. Open the sample application
2. Click on the button
3. Validate that the first DateTime Cell get the data with the seconds information

### Checklist
* [x] tested locally
* [x] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/CellData.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/CellData.ts
@@ -19,7 +19,9 @@ namespace Providers.DataGrid.Wijmo.Feature {
 
 		public setCellData(rowNumber: number, column: OSFramework.DataGrid.Column.IColumn, value: string): void {
 			if (column.columnType === OSFramework.DataGrid.Enum.ColumnType.DateTime) {
-				value = this._grid.dataSource.trimSecondsFromDate(value);
+				if (column.config.format.trim().toLowerCase().includes('s')) {
+					value = this._grid.dataSource.trimSecondsFromDate(value);
+				}
 			}
 
 			this._grid.provider.setCellData(rowNumber, column.provider.index, value, true);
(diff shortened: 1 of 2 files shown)


src/Providers/DataGrid/Wijmo/Features/CellData.ts
```diff
@@ -19,7 +19,9 @@ namespace Providers.DataGrid.Wijmo.Feature {
 
 		public setCellData(rowNumber: number, column: OSFramework.DataGrid.Column.IColumn, value: string): void {
 			if (column.columnType === OSFramework.DataGrid.Enum.ColumnType.DateTime) {
-				value = this._grid.dataSource.trimSecondsFromDate(value);
+				if (column.config.format.trim().toLowerCase().includes('s')) {
```
rugoncalves: Shouldn't it be "ss"? `MM-dd-yyyy hh:mm:ss → "07-09-2021 12:34:02"`
Just to be easier to match both sides? open question...

rugoncalves (dismissed):
Looks like a good solution!

src/Providers/DataGrid/Wijmo/Features/CellData.ts
os-davidlourenco: I set as 's' because the user can set in the format to only use the first digit of seconds. But also not sure if makes sense

rugoncalves: it sounds far-fetched, but the world is full of non-sense 😂 
Approved ✅

rugoncalves (approved):
Looks good!

PR #508, opened 2026-06-06
ROU-12848: Fix Broken ARIA reference on Data Grid column headers

gnbm:
This PR fixes the "Broken ARIA reference" accessibility errors (reported by WAVE) on Data Grid column headers and filter buttons, and unblocks CI by resolving a pre-existing `npm install` dependency conflict.

### What was happening

**Accessibility (primary)**
* Every filterable column header cell and its filter ("funnel") button carried an `aria-describedby` attribute pointing to a non-existent element (e.g. `aria-describedby="0.lhfuia0d7am"`), producing WAVE "Broken ARIA reference" errors (2 per filterable column).
* Root cause: the wrapper stored the OutSystems column `uniqueId` in Wijmo's `Column.describedById` property purely as a side-channel for column lookup. Wijmo renders `describedById` as a real `aria-describedby` attribute, but no matching description element was ever created — so the IDREF dangled.
* axe-core and Lighthouse stayed green because they don't validate `aria-describedby` targets; plain Wijmo FlexGrid (same version) emits no `aria-describedby`, confirming the defect lived in the wrapper, not Wijmo.

**CI / dependencies**
* `npm install` failed in CI with `ERESOLVE`. A prior automated bump (`af11b3b`) raised `typedoc` to `^0.28.19` (and `typedoc-umlclass` to `^0.10.2`) without a compatible TypeScript. `typedoc@0.28` peer-requires `typescript@>=5.0`, but the project is pinned to `typescript@^4.9.5`; it also breaks `typedoc-plugin-merge-modules@4.1.0` (peer `typedoc 0.23.x||0.24.x`).
* Because `package-lock.json` is gitignored, CI resolves fresh from `package.json` on every install, so this conflict broke every PR (pre-existing on `dev`, surfaced — not caused — by this PR).

### What was done

**Accessibility fix**
* Stopped passing the `uniqueId` to Wijmo's `describedById` (removed from `ColumnConfig.getProviderConfig()` and the `IColumnProviderConfigs` interface), so no `aria-describedby` attribute is emitted.
* Moved the `uniqueId` to Wijmo's documented `Column.name` property — a DOM-inert identifier intended for distinguishing columns (including ones that share a binding). It is set via `name: this.uniqueId` in `ColumnConfig.getProviderConfig()` and read back through `column.name` in `ClickEvent` and `DirtyMark`, keeping the existing `binding` fallback for autogenerated columns. Column-to-OutSystems-column mapping is behaviorally identical.
* Note: an earlier iteration stored the id on a private expando; following PR review (it relied on undocumented Wijmo behavior, like past `RowMetadata` issues) it was changed to the supported `Column.name` property.
* Fixed two pre-existing comment typos in touched files (`Cleanning` → `Cleaning`, `not need` → `no need`).

**TypeDoc rollback (CI fix)**
* Reverted `typedoc` `^0.28.19` → `^0.23.28` and `typedoc-umlclass` `^0.10.2` → `^0.7.1` (the last versions compatible with TypeScript 4.9.5 and `typedoc-plugin-merge-modules@4.1.0`), restoring a peer-coherent set and a clean `npm install`. `typedoc` is dev-only docs tooling — no runtime impact. Kept TypeScript at `^4.9.5`; the TypeScript 5.x upgrade is tracked separately under ROU-11724.

**Dependency bump (bundled)**
* postcss → `^8.5.15`.

**Documentation**
* Added `ADR-0005` (Broken ARIA reference fix) and `ADR-0006` (TypeDoc rollback, referencing ROU-11724) under `docs/adr/`, and backfilled the missing `ADR-0004` entry in the ADR log.

### Test Steps
1. Open the sample page
2. Run WAVE (or the DevTools IDREF audit) — confirm **0** "Broken ARIA reference" errors (was 12).
3. Regression: click a data cell and confirm the OnCellClick event fires with the correct column; edit a cell and confirm the dirty-mark styling applies.
4. Confirm filtering, sorting, reorder, grouping, column picker and Excel export still work (these touch column identity, now that `name` is populated).
5. CI: confirm `npm install` resolves with no `ERESOLVE` and `build-n-test` passes. (`npm run docs` still generates with the reverted TypeDoc.)

### Screenshots
(to be added — before/after WAVE run on Screen2)

### Checklist
* [x] tested locally
* [x] documented the code


package.json
OS-giulianasilva, line 1: @gnbm, as this downgrade was necessary because of the `TypeScript` version we are using, I would suggest to creating a task to update the `TypeScript`, update these package that were downgraded or any other needed and remove the `package-lock.json` from the `.gitignore`.

src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts
```diff
@@ -44,6 +44,11 @@ namespace OSFramework.DataGrid.Configuration.Column {
 		}
 
 		public getProviderConfig(): DataGrid.Types.IColumnProviderConfigs {
+			// NOTE: the column's uniqueId is intentionally NOT passed to Wijmo's `describedById`
```
OS-giulianasilva: @gnbm , I am not feeling secure about the use of the `__osUniqueId` property, just because this is not officially supported by Wijmo and I remember that we had issues with the `RowMetadata` in the past because of this. 

I would suggest for us to use the [Column's name prop](https://developer.mescius.com/wijmo/api/classes/Wijmo_Grid.Column.html#name) to set the `uniqueId` instead.

OS-giulianasilva (changes requested):

package.json
gnbm: @OS-giulianasilva we already have this task `ROU-11724` 👍🏻

OS-giulianasilva (approved):

PR #509, opened 2026-06-08
ROU-10907: Remove Lodash dependency

rugoncalves:
This PR is for ROU-10907 — removing the Lodash dependency from the TypeScript project and replacing all usages with native ES2020 equivalents or small typed utility functions.

### What was happening
  * The project depended on `lodash@4.18.1` for 9 utility functions spread across 23 call sites in 13 files.
  * Lodash has a history of high-severity CVEs (prototype pollution, ReDoS), requiring reactive version-pinning
  work on each new advisory.
  * All functions in use have direct native ES2020 equivalents or are trivially implementable without a
  third-party library.

### What was done
* Removed `lodash` and `@types/lodash` from `package.json` `devDependencies`.
* Added four typed utility functions to `OSFramework.DataGrid.Helper` (`src/OSFramework/DataGrid/Helper/Utils.ts`): 
   * `GetByPath` (dot-notation path accessor)
   * `SetByPath` (dot-notation path setter)
   * `DeepClone` (backed by `structuredClone`, correctly handles `Date` objects and `undefined` properties)
   * `Omit`
  * Replaced all 23 Lodash call sites across 13 files with native JavaScript or the new utilities - full replacement map in [ADR-0005](../docs/adr/ADR-0005-Remove-Lodash-Dependency.md).
  * Updated TypeScript target from `ES2019` to `ES2020`, enabling native emission of `?.` optional chaining and `??` nullish coalescing - used to simplify `GetByPath` and `SetByPath`.
  * Corrected pre-existing incorrect `JSON[]` type annotations (the global `JSON` object interface, not a data type) to `object[]` on `addRow` in `IDataSource.ts`, `AbstractDataSource.ts`, and `ProviderDataSource.ts`.
  * Added ADR-0005 documenting the decision, rationale, and full replacement map.

### Test Steps
  1. Run full regression pipeline

### Screenshots
  N/A — no visual or behavioural changes.

### Checklist
  * [x] tested locally
  * [x] documented the code (ADR-0005 added to `docs/adr/`)
  * [x] requires changes in OutSystems — the Lodash `.js` resource must be removed from the OutSystems DataGrid
  module (the runtime global `_` is no longer referenced).
  * [ ] requires new sample page in OutSystems


package.json
```diff
@@ -58,5 +56,9 @@
     "typedoc-umlclass": "^0.10.2",
     "typescript": "^4.9.5",
     "wijmo": "^5.20261.50"
+  },
+  "volta": {
+    "node": "24.13.1",
+    "npm": "11.10.1"
   }
```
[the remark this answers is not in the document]
rugoncalves: Accepted.

src/OSFramework/DataGrid/Helper/Utils.ts
```diff
@@ -6,11 +6,41 @@ namespace OSFramework.DataGrid.Helper {
...
+
+	export function SetByPath(obj: Record<string, unknown>, path: string, value: unknown): void {
+		const keys = path.split('.');
+		let current: Record<string, unknown> = obj;
+		for (let i = 0; i < keys.length - 1; i++) {
+			current[keys[i]] = current[keys[i]] ?? {};
+			current = current[keys[i]] as Record<string, unknown>;
+		}
+		current[keys[keys.length - 1]] = value;
+	}
```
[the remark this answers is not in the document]
rugoncalves: accepted.

src/OSFramework/DataGrid/Helper/Utils.ts
```diff
@@ -6,11 +6,41 @@
...
+
+	export function SetByPath(obj: Record<string, unknown>, path: string, value: unknown): void {
+		const keys = path.split('.');
+		let current: Record<string, unknown> = obj;
+		for (let i = 0; i < keys.length - 1; i++) {
+			current[keys[i]] = current[keys[i]] ?? {};
+			current = current[keys[i]] as Record<string, unknown>;
+		}
+		current[keys[keys.length - 1]] = value;
```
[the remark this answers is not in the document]
rugoncalves: accepted.

src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
```diff
@@ -166,7 +166,7 @@ namespace OSFramework.DataGrid.Grid {
 
 		private _getRowByKey(key: string) {
 			return this._ds.find((item) => {
-				return _.get(item, this.parentGrid.config.keyBinding).toString() === key;
+				return String(Helper.GetByPath(item, this.parentGrid.config.keyBinding)) === key;
```
os-davidlourenco, line 169: I have a question regarding this.
In the GetByPath definition, the comments ensure that it is possible to return undefined in some cases. In those cases, I'm not sure if this will not be an issue since the `.toString` returns undefined, but the String(...) returns "undefined" (as a string). 
Not sure if this applies in this case, but it could be a point of failure if it does.

src/OSFramework/DataGrid/Helper/Utils.ts
```diff
@@ -1,16 +1,88 @@
...
 	 * @param data Array to be split
 	 * @param callback Callback to be executed on splitted arrays
 	 * @param chunkSize Size of splitted array
 	 */
-	export function BatchArray(data: Array<string | number | JSON>, callback: (part) => void, chunkSize = 10): void {
+	export function BatchArray(data: object[], callback: (part: object[]) => void, chunkSize = 10): void {
 		if (!Array.isArray(data)) throw new Error('An array must be passed');
 
-		const chunk = _.chunk(data, chunkSize);
+		const chunk = Array.from({ length: Math.ceil(data.length / chunkSize) }, (_, i) =>
```
os-davidlourenco, line 16: If chunkSize is 0, this could be a problem since we would try to divide by 0, creating an infinite array

src/OSFramework/DataGrid/Helper/Utils.ts
```diff
@@ -1,16 +1,88 @@
...
+	 * Creates a deep clone of the given object.
+	 * Uses structuredClone if available, otherwise falls back to JSON methods.
+	 *
+	 * @param obj The object to clone
+	 * @returns A deep clone of the object
+	 */
+	export function DeepClone<T>(obj: T): T {
+		try {
+			// structuredClone is available in modern environments and handles more
+			// complex cases than JSON methods (like functions, Dates, Maps, Sets, etc.)
```
os-davidlourenco: The structureClone() doesn't clone functions as far as I'm aware. Not sure if this would be an issue here, but at least comment should be updated

src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
rugoncalves: Since we're comparing to key, that doesn't sound as a problem. Do you think it's worth to:
```
const result = Helper.GetByPath(item, this.parentGrid.config.keyBinding);
return !!result && String(result) === result;
```

src/OSFramework/DataGrid/Helper/Utils.ts
rugoncalves: Protected the code.

src/OSFramework/DataGrid/Helper/Utils.ts
rugoncalves: Updated comment for clarity.

src/OSFramework/DataGrid/Grid/AbstractDataSource.ts
os-davidlourenco: You are right, it shouldn't be a problem. I would left as it is since it's more concise

os-davidlourenco (dismissed):

src/OSFramework/DataGrid/Helper/Utils.ts
```diff
@@ -1,16 +1,93 @@
...
 	/**
 	 * Receives an array, splits it into smaller arrays and executes a callback.
 	 * @param data Array to be split
 	 * @param callback Callback to be executed on splitted arrays
 	 * @param chunkSize Size of splitted array
 	 */
-	export function BatchArray(data: Array<string | number | JSON>, callback: (part) => void, chunkSize = 10): void {
+	export function BatchArray(data: object[], callback: (part: object[]) => void, chunkSize = 10): void {
 		if (!Array.isArray(data)) throw new Error('An array must be passed');
+		if (chunkSize <= 0) throw new Error('Chunk size must be a positive integer');
```
OS-giulianasilva:
```suggestion
		if (chunkSize <== 0) throw new Error('Chunk size must be a positive integer');
```

OS-giulianasilva (changes requested):

src/OSFramework/DataGrid/Configuration/Column/ColumnConfigConditionalFormat.ts
```diff
@@ -6,8 +6,6 @@ namespace OSFramework.DataGrid.Configuration.Column {
 	 * Defines the configuration for Calculated Columns
 	 */
 	export class ColumnConfigConditionalFormat extends ColumnConfig {
-		public conditionalFormat: Array<OSStructure.ConditionalFormat>;
```
OS-giulianasilva, line 9: Why was this removed?

src/OSFramework/DataGrid/Configuration/Column/EditorConfigDate.ts
```diff
@@ -5,7 +5,6 @@ namespace OSFramework.DataGrid.Configuration.Column {
 	 */
 	export class EditorConfigDate extends AbstractEditorConfig {
 		public defaultFormat: string;
-		public format: string;
```
OS-giulianasilva, line 8: Why was this removed?

rugoncalves: Because the property exists in the base class: https://github.com/OutSystems/outsystems-datagrid/blob/94bc0cfc98b9579315483ba08e0794060f11e590/src/OSFramework/DataGrid/Configuration/Column/AbstractEditorConfig.ts#L13

src/OSFramework/DataGrid/Configuration/Column/ColumnConfigConditionalFormat.ts
rugoncalves: Because the property exists in the base class: https://github.com/OutSystems/outsystems-datagrid/blob/94bc0cfc98b9579315483ba08e0794060f11e590/src/OSFramework/DataGrid/Configuration/Column/ColumnConfig.ts#L20

OS-giulianasilva (dismissed):

src/OSFramework/DataGrid/Helper/Utils.ts
rugoncalves: This change is actually invalid. I've undo the suggestion and change. Thank you!

OS-giulianasilva (approved):

PR #511, opened 2026-06-24
ROU-12864: Add new parameter to SetCellData client action

os-davidlourenco:
This PR is to add a new parameter to SetCellData Client Action to consider the value as a key

### What was happening
* The SetCellData Client Action was considering that the value could be the key or the value.
* This made the internal Wijmo logic look for a key with the value passed as a parameter.
* In some cases, the key could be used as a value and trigger a search for other options

### What was done
* A new parameter was added to the API function
* The internal call is now done to a different function with the same behaviour that allow to pass a parameter to directly consider the value passed as an `option key`

### Test Steps
1. Open the sample application
2. Click on the "Set 3" button
3. Validate that the text "Option3" is presented

### Checklist
* [ ] tested locally
* [ ] documented the code

### DIFF
--- a/src/OutSystems/GridAPI/Cell.ts
+++ b/src/OutSystems/GridAPI/Cell.ts
@@ -106,7 +106,8 @@ namespace OutSystems.GridAPI.Cells {
 		// eslint-disable-next-line
 		value: any,
 		showDirtyMark = true,
-		triggerOnCellValueChange = true
+		triggerOnCellValueChange = true,
+		valueIsKey = false
 	): string {
 		Performance.SetMark('Cells.setCellData');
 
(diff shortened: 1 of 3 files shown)


src/Providers/DataGrid/Wijmo/Features/CellData.ts
```diff
@@ -17,14 +17,14 @@ namespace Providers.DataGrid.Wijmo.Feature {
...
-		public setCellData(rowNumber: number, column: OSFramework.DataGrid.Column.IColumn, value: string): void {
+		public setCellData(rowNumber: number, column: OSFramework.DataGrid.Column.IColumn, value: string, valueIsKey: boolean): void {
 			if (column.columnType === OSFramework.DataGrid.Enum.ColumnType.DateTime) {
 				if (!column.config.format.trim().toLowerCase().includes('s')) {
 					value = this._grid.dataSource.trimSecondsFromDate(value);
 				}
 			}
 
-			this._grid.provider.setCellData(rowNumber, column.provider.index, value, true);
+			this._grid.provider.cells.setCellData(rowNumber, column.provider.index, value, true, true, valueIsKey);
```
rugoncalves, line 28: Add a comment here in the code.

src/OutSystems/GridAPI/Cell.ts
rugoncalves, line 100: Update method documentation.

rugoncalves (commented):
Looking good! Minor documentation additions required!

rugoncalves (approved):

PR #516, opened 2026-06-30
ROU-12831: Match Wijmo columns by name instead of index or binding

OS-giulianasilva:
This PR is for making the Data Grid resolve Wijmo columns by their stable name identifier instead of the column index or binding.

### What was happening
* Several features mapped a Wijmo column back to its OutSystems column by comparing the column index or binding. The index is explicitly discouraged by the vendor and breaks when columns are reordered, hidden, frozen or grouped (group and data columns can even share an index), which could fire OnCellValueChange, place validation marks, or apply conditional formatting against the wrong column.

### What was done
* Added a single getColumnByProvider helper on the Wijmo grid wrapper that resolves an OutSystems column from a Wijmo column by the stable name (uniqueId), falling back to binding for autogenerated columns that have no name, and excluding Group columns.
* ValidationMark: replaced the 3 index-based lookups (cell edit, undo, redo) with the helper.
* ConditionalFormat: matched the provider column by name/uniqueId (with a binding fallback for autogenerated columns) in the row and aggregate updates, addressed cells by stable id, and hoisted the per-column rule lookup out of the row loop.
* ContextMenu: resolved the hit-test column via the helper and reset the column context up front to avoid reporting a stale column.
* ColumnPicker: resolved the list item column via the helper.
* GroupColumn: matched the child column by name, falling back to binding when the provider column has no name.
* DirtyMark: saveOriginalValue now identifies the cell's column by its stable identifier (uniqueId, falling back to binding) instead of a positional index; updated its callers accordingly (SetCellData, the cell-edit handler, the undo stack, and the dropdown column).
* DropdownColumn: cast the grid to IGridWijmo when building the child edit action, since IGridWijmo is no longer structurally empty after gaining the helper.

### Test Steps
1. Open a grid with reordered columns and at least one grouped column.
2. Edit a cell, paste a value, then undo and redo; confirm OnCellValueChange fires for the correct column and the dirty mark appears on the correct cell each time.
3. Confirm mandatory-cell validation marks land on the correct cell and column.
4. Confirm per-row and aggregate conditional formatting apply to the correct column.
5. Repeat with a hidden and a frozen column.
6. Right-click different columns (including a group column header and an empty area) and confirm the context menu reports the correct column (or none).
7. On an autogenerated grid (no explicit column blocks), confirm dirty marks, validation marks, conditional formatting and the context menu still resolve the correct column.

### Screenshots
(prefer animated gif)

### Checklist
* [ ] tested locally
* [ ] documented the code


src/Providers/DataGrid/Wijmo/Features/ConditionalFormat.ts
```diff
@@ -268,23 +268,29 @@ namespace Providers.DataGrid.Wijmo.Feature {
...
 				);
 
-				// Execute the rule
+				// Execute the rule. `col` is the column's stable string identifier, consistent with
+				// `_updateRows`; `ConditionExecuter` resolves it via `provider.getColumn`.
 				this._mappedRulesAggregate.get(column.config.binding).execute(
 					aggregateValue,
 					{
-						col: colIndex,
+						col: column.config.uniqueId || column.config.binding,
```
OS-giulianasilva, line 295: col is used in `FlexGrid.getCellData()` inside the `ConditionExecuter.execute()`. The `FlexGrid.getCellData()` accepts `binding`, `name`, and `index`. Changed it to use the `name`/`uniqueId `or `binding`, for consistency.

src/Providers/DataGrid/Wijmo/Features/ConditionalFormat.ts
```diff
@@ -268,23 +268,29 @@ namespace Providers.DataGrid.Wijmo.Feature {
 
 			// Iterate columns in order to get aggregate cell values
 			columnsAggregate.forEach((column) => {
-				const colIndex = this._grid.provider.columns.find((x) => x.binding === column.provider.binding).index;
+				// Match the Wijmo provider column by the stable `name` identifier (which holds the
+				// OSFramework uniqueId), falling back to `binding` for autogenerated columns (no `name`).
+				// The resolved `.index` is read fresh.
+				const providerColumn = this._grid.provider.columns.find(
+					(x) => x.name === column.uniqueId || (!x.name && x.binding === column.provider.binding)
```
rugoncalves: Use a more readable variable name (instead of `x`).

OS-giulianasilva: done!

rugoncalves (approved):
What a beautiful change!

PR #525, opened 2026-07-22
ROU-12917: Refactor cell edit handling to use shared evaluation logic for manual edits and pastes

OS-giulianasilva:
This PR is for fixing the ValidationMark handler so a pasted cell's original value is read from the correct event property.

### What was happening
* The ValidationMark feature bound a single handler to both the `cellEditEnded` and `pastedCell` Wijmo events.
* That handler read the cell's original value from `e.previousData`, which is only populated during the edit lifecycle and is `undefined` on paste.
* Because the original value was `undefined` on paste, the "value actually changed" gate evaluated incorrectly, producing wrong validation marks and `OnCellValueChange` behaviour when pasting into columns filtered by ParentBinding.

### What was done
* Extracted the shared change-gate and trigger logic into a private `_evaluateCellChange` method that receives the original value as a parameter.
* `_cellEditEndedHandler` is now a thin wrapper that keeps passing `e.previousData`, so manual edit behaviour is unchanged.
* Added a dedicated `_pastedCellHandler` that passes `e.data`, which holds the cell's original value before the paste.
* Updated `build()` to bind the `pastedCell` event to the new paste handler instead of the edit handler.

### Test Steps
1. Open a grid using the Column_DropdownDependency sample with a mandatory, ParentBinding-dependent column.
2. Copy a child value that is valid under one parent and paste it into a row whose parent does not allow it.
3. Confirm the target cell shows the correct validation mark and the mandatory tooltip.
4. Manually edit a cell, then undo and redo, and confirm validation behaviour is unchanged.

### Screenshots
(prefer animated gif)

### Checklist
* [ ] tested locally
* [ ] documented the code

### DIFF
--- a/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
+++ b/src/Providers/DataGrid/Wijmo/Features/ValidationMark.ts
@@ -33,25 +33,8 @@ namespace Providers.DataGrid.Wijmo.Feature {
 		 * Handler for the CellEditEnded.
 		 */
 		private _cellEditEndedHandler(s: wijmo.grid.FlexGrid, e: wijmo.grid.CellEditEndingEventArgs): void {
-			// get the new value
-			const newValue = s.getCellData(e.row, e.col, false);
-			const previousValue = e.previousData;
-
-			const isNewValue = previousValue !== newValue && previousValue?.toString() !== newValue?.toString();
-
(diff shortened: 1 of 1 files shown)


JoaoFerreira-FrontEnd (approved):
LGTM - nice job

